PR #1, opened 2016-05-30
Test on the GNU toolchain.

frewsxcv:


steffengy:
superseded by https://github.com/sfackler/schannel-rs/pull/2 .

While the major work is done somewhere else, it doesn't really make sense
to touch this inactive state here, might only make things harder in the future.

frewsxcv:
@steffengy While you're around, do you have any ideas what might be causing the failures in https://github.com/sfackler/schannel-rs/pull/3 for 32bit machines?

steffengy:
@frewsxcv fixed it, filing a PR.

PR #3, opened 2016-07-27
Client side of rewrite

sfackler:
This only contains client-side support currently, but server support should be add-able in a backwards compatible manner in the future.


steffengy:
looks good to me so far, thank you :+1:

PR #5, opened 2016-07-28
Add a MidHandshakeTlsStream

alexcrichton:
Along the lines of `security-framework`'s version of this (and that being added
to [rust-openssl][openssl]) this adds a `MidHandshakeTlsStream` to allow
applications to concretely know when a handshake has been completed (e.g.
validation has all been cleared).

[openssl]: https://github.com/sfackler/rust-openssl/pull/424

### DIFF
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -63,13 +63,19 @@ impl Builder {
     }
 
     /// Initializes a new TLS session.
-    pub fn initialize<S>(&self, mut cred: SchannelCred, stream: S) -> io::Result<TlsStream<S>>
+    pub fn initialize<S>(&self,
+                         mut cred: SchannelCred,
+                         stream: S)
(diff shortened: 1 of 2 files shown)


src/tls_stream.rs
```diff
@@ -123,6 +125,23 @@ pub struct TlsStream<S> {
     out_buf: Cursor<Vec<u8>>,
 }
 
+/// A failure which can happen during the `Builder::initialize` phase, either an
+/// I/O error or an intermediate stream which has not completed its handshake.
+#[derive(Debug)]
+pub enum HandshakeError<S> {
```
sfackler: This should probably implement Error

sfackler:
One test failed. It may just need to be deleted since it covers an error case that doesn't exist after this change.

alexcrichton:
Hm yeah, there's a few other updates I want to do on the accept branch I'm working on as well, but I think I'll PR it all at once, so closing for now.

PR #6, opened 2016-07-28
Move the SChannelCred Builder to &mut self

alexcrichton:
... instead of threading through ownership. This matches builders in the
standard library like `Command`.


sfackler:
Probably worth making sure that there aren't any future configuration options that would require an ownership transfer.

alexcrichton:
The standard library's `Command` has that (via stdin/stdout/stderr) and that's just done via `Option` and ownership transfer. It may mean though that the "build" method should take `&mut self`

PR #7, opened 2016-07-29
Implementation of accepting TLS connections

alexcrichton:
These are a few commits which lead up to the conclusion of accepting server-side TLS connections. There's a bit more here than I initially anticipated, so a close eye would definitely be appreciated! The highlights are:

* TLS streams can call `AcceptSecurityContext` instead of just `InitializeSecurityContextW` to kick off the process of accepting a stream.
* `CertStore`, `CertContext`, and `CtlContext` were all exposed to expose management of keys and such.
* Builders in a few locations got arguments to each of these new contexts.
* The contexts were expanded with a suite of inspection/creation methods to manage them
* A test was added which spins up a server, accepts a connection, communicates a little, and then exits.

The hard part about accepting connections on the server side is that we need the client to trust our custom certificate. Currently the only way I know to do this is to add a certificate to the global trusted root certificate store, so that's what this recommends! I added a script `test/create.sh` to create a custom CA, sign a certificate for `localhost`, and then delete the CA key. That way those who want to run tests can add the custom CA. Each test should also be able to detect whether our custom CA is in the system store, and also provides the option to opt-out via environment variables.

It's my intention that the tests run on AppVeyor, but I haven't tested that much, so I'll probably need some tweaks here and there for that.


alexcrichton:
Ah and this includes the refactoring in https://github.com/steffengy/schannel-rs/pull/5 of a `MidHandshakeTlsStream`

appveyor.yml
```diff
@@ -7,7 +7,7 @@ environment:
   access_token:
     secure: ZxcrtxQXwszRYNN6c1ZIagczEqzmQQZeYHY58izcmF0jdq/cptxJvFUoVxDmnoqj
 install:
-  - ps: New-SelfSignedCertificate -DnsName localhost -CertStoreLocation cert:Localmachine\My
+  - ps: Import-Certificate -FilePath test/schannel-ca.crt -CertStoreLocation 'Cert:\CurrentUser\Root' -Verbose
```
steffengy: This will be very annoying...
UAC doesn't allow that, since no "arbitrary" program should be able to touch the root certificate store.
So with any solution I know, the confirmation dialog pops up.

alexcrichton: Hm yeah I was looking into this. Apparently appveyor [has a semi-solution for this](http://www.appveyor.com/docs/deployment/local#installing-self-signed-ssl-certificate-to-a-website) though? I'll try that and see if it works better.

steffengy: funny fact installing to ```Cert:\LocalMachine\Root``` works, is that a solution?

alexcrichton: Let's find out! Certainly easier than adding a new script...

alexcrichton:
Oh hey looks like @steffengy your suggestion worked! CI is now passing.

src/cert_context.rs
```diff
@@ -1,13 +1,138 @@
...
+    ///
+    /// The `<SIGNATURE>` value identifies the CNG public key
+    /// algorithm. The `<HASH>` value identifies the CNG hash algorithm.
+    ///
+    /// Common examples are:
+    ///
+    /// * `RSA/SHA1`
+    /// * `RSA/SHA256`
+    /// * `ECDSA/SHA256`
+    pub fn algorithm_hash(&self) -> io::Result<String> {
```
sfackler: Not sure about this name. `hash_algorithm` reads a bit better, but doesn't cover the key algorithm. If we wanted to follow winapi naming, `sign_hash_algorithms` might be that?

src/cert_store.rs
```diff
@@ -63,11 +70,95 @@ impl CertStore {
...
+                self.cur = Some(next.clone());
+                Some(next)
+            }
+        }
+    }
+}
+
+/// An in-memory store of certificates and CTLs, created by `CertStore::memory`
+/// and can be converted into a `CertStore`.
+#[derive(Clone)]
```
sfackler: Before we publish this, I'd like to figure out naming of this thing. It's really a builder type and I'm not sure if `Memory` is the right way to name it. On the other hand, there will be more builder types added over time (like one for a store that acts as the union of a set of stores) so it should be descriptive of what it is.

src/ctl_context.rs
```diff
@@ -40,6 +44,7 @@ impl Inner<winapi::PCCTL_CONTEXT> for CtlContext {
 }
 
 impl CtlContext {
+    /// Returns a builder reader to create an encode `CtlContext`.
```
sfackler: `encode`

src/ctl_context.rs
```diff
@@ -113,7 +127,9 @@ impl Builder {
                                                         ptr::null_mut(),
                                                         &mut size);
             if res == winapi::FALSE {
-                return Err(io::Error::last_os_error());
+                let err = io::Error::last_os_error();
+                debug!("crypt and sign error: {:?}", err);
```
sfackler: Should probably not be here.

src/ctl_context.rs
```diff
@@ -125,7 +141,9 @@ impl Builder {
                                                         encoded.as_mut_ptr() as *mut winapi::BYTE,
                                                         &mut size);
             if res == winapi::FALSE {
-                return Err(io::Error::last_os_error());
+                let err = io::Error::last_os_error();
+                debug!("crypt and sign error with bytes: {:?}", err);
```
sfackler: ^

src/ctl_context.rs
```diff
@@ -48,24 +53,33 @@ impl CtlContext {
...
 }
 
+/// Used to build an encoded `CtlContext` which can be added to a `Memory` store
+/// to get back the actual `CtlContext`.
 pub struct Builder {
     certificates: Vec<CertContext>,
     usages: Vec<Vec<u8>>,
 }
 
 impl Builder {
```
sfackler: We'll need to allow signers to be added before it makes sense for this to be public.

src/schannel_cred.rs
```diff
@@ -167,10 +169,24 @@ impl Builder {
...
+    /// Clients often do not call this function and either depend on Schannel to
+    /// find an appropriate certificate or create a certificate later if needed.
+    pub fn cert(&mut self, cx: CertContext) -> &mut Builder {
+        self.certs.push(cx);
+        self
+    }
+
     /// Creates a new `SchannelCred`.
-    pub fn acquire(&self, direction: Direction) -> io::Result<SchannelCred> {
+    pub fn acquire(&mut self, direction: Direction) -> io::Result<SchannelCred> {
```
sfackler: why mut?

src/security_context.rs
```diff
@@ -67,37 +66,48 @@ impl SecurityContext {
                                                       &mut ctxt,
                                                       &mut outbuf_desc,
                                                       &mut attributes,
-                                                      ptr::null_mut()) {
-                winapi::SEC_I_CONTINUE_NEEDED => Ok((SecurityContext(ctxt), ContextBuffer(outbuf))),
-                err => Err(io::Error::from_raw_os_error(err as i32)),
+                                                      ptr::null_mut()) {
+                winapi::SEC_I_CONTINUE_NEEDED => {
+                    debug!("initialize incomplete");
```
sfackler: ^

src/security_context.rs
```diff
@@ -67,37 +66,48 @@ impl SecurityContext {
...
-                winapi::SEC_I_CONTINUE_NEEDED => Ok((SecurityContext(ctxt), ContextBuffer(outbuf))),
-                err => Err(io::Error::from_raw_os_error(err as i32)),
+                                                      ptr::null_mut()) {
+                winapi::SEC_I_CONTINUE_NEEDED => {
+                    debug!("initialize incomplete");
+                    Ok((SecurityContext(ctxt), Some(ContextBuffer(outbuf[0]))))
+                }
+                err => {
+                    let err = io::Error::from_raw_os_error(err as i32);
+                    debug!("initialize error: {:?}", err);
```
sfackler: ^

src/security_context.rs
```diff
@@ -67,37 +66,48 @@ impl SecurityContext {
...
+            let mut stream_sizes = mem::zeroed();
             let status = secur32::QueryContextAttributesW(&mut self.0,
                                                           winapi::SECPKG_ATTR_STREAM_SIZES,
                                                           &mut stream_sizes as *mut _ as *mut _);
             if status == winapi::SEC_E_OK {
                 Ok(stream_sizes)
             } else {
-                Err(io::Error::from_raw_os_error(status as i32))
+                let err = io::Error::from_raw_os_error(status as i32);
+                debug!("remote cert error {:?}", err);
```
sfackler: ^

src/security_context.rs
```diff
@@ -67,37 +66,48 @@ impl SecurityContext {
...
+            let mut cert_context = mem::zeroed();
             let status = secur32::QueryContextAttributesW(&mut self.0,
                                                           winapi::SECPKG_ATTR_REMOTE_CERT_CONTEXT,
                                                           &mut cert_context as *mut _ as *mut _);
             if status == winapi::SEC_E_OK {
                 Ok(CertContext::from_inner(cert_context))
             } else {
-                Err(io::Error::from_raw_os_error(status as i32))
+                let err = io::Error::from_raw_os_error(status as i32);
+                debug!("remote cert error {:?}", err);
```
sfackler: ^

src/test.rs
```diff
@@ -144,13 +161,163 @@ fn validation_failure_is_permanent() {
...
+fn can_trust_localhost_der() -> bool {
+    if env::var("SCHANNEL_RS_SKIP_SERVER_TESTS").is_ok() {
+        return false
+    }
+    let cert = include_bytes!("../test/schannel-ca.der");
+    let cert = CertContext::new(cert).unwrap();
+    let hash_to_find = cert.signature_hash().unwrap();
+    let mut root = CertStore::system("Root").unwrap();
+    let mut my = CertStore::system("My").unwrap();
+    for cert in root.iter().chain(my.iter()) {
```
sfackler: Probably worth binding the APIs that allow schannel do this lookup for us.

src/tls_stream.rs
```diff
@@ -56,25 +59,49 @@ impl Builder {
...
     }
 
-    #[allow(dead_code)]
-    /* pub */ fn cert_store(&mut self, cert_store: CertStore) -> &mut Builder {
+    /// Specifies a custom certificate store which is later used when validating
+    /// a server's certificate.
+    ///
+    /// This option is only used for client connections and is used to construct
+    /// the certificate chain which the server's certificate is validated
+    /// against.
```
sfackler: We'll probably want to note here that the certificates in this store are not implicitly trusted - it certainly surprised me.

src/tls_stream.rs
```diff
@@ -56,25 +59,49 @@ impl Builder {
...
+    pub fn cert_store(&mut self, cert_store: CertStore) -> &mut Builder {
         self.cert_store = Some(cert_store);
         self
     }
 
+    /// Indicates whether this builder will be accepting a connection or not.
+    ///
+    /// For servers this flag should be set to `true`, and for clients no action
+    /// need be taken as this option is `false` by default.
+    pub fn accept(&mut self, accept: bool) -> &mut Builder {
```
sfackler: I'd rather make this another "terminal" method like `initialize`, or store this state in the `SchannelCred` and figure it out for the user.

src/tls_stream.rs
```diff
@@ -123,6 +148,23 @@ pub struct TlsStream<S> {
     out_buf: Cursor<Vec<u8>>,
 }
 
+/// A failure which can happen during the `Builder::initialize` phase, either an
+/// I/O error or an intermediate stream which has not completed its handshake.
+#[derive(Debug)]
+pub enum HandshakeError<S> {
```
sfackler: This should implement Error

src/tls_stream.rs
```diff
@@ -193,62 +231,57 @@ impl<S> TlsStream<S>
...
-                ulVersion: winapi::SECBUFFER_VERSION,
-                cBuffers: 3,
-                pBuffers: outbufs.as_mut_ptr(),
-            };
+            let pos = self.enc_in.position() as usize;
+            let mut inbufs = [secbuf(winapi::SECBUFFER_TOKEN,
+                                     Some(&mut self.enc_in.get_mut()[..pos])),
+                              secbuf(winapi::SECBUFFER_EMPTY, None)];
+            let mut inbuf_desc = secbuf_desc(&mut inbufs);
+            debug!("initializing/accepting with {} input bytes", pos);
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -193,62 +231,57 @@ impl<S> TlsStream<S>
...
-                                                             &mut outbuf_desc,
-                                                             &mut attributes,
-                                                             ptr::null_mut());
+            let status = if self.accept {
+                let ptr = if self.accept_first {
+                    ptr::null_mut()
+                } else {
+                    self.context.get_mut()
+                };
+                debug!("accept({:p})", self.context.get_mut());
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -193,62 +231,57 @@ impl<S> TlsStream<S>
...
-                                                             0,
-                                                             0,
-                                                             &mut inbuf_desc,
-                                                             0,
-                                                             ptr::null_mut(),
-                                                             &mut outbuf_desc,
-                                                             &mut attributes,
-                                                             ptr::null_mut());
+            let status = if self.accept {
+                let ptr = if self.accept_first {
```
sfackler: Is it possible to avoid this weirdness and always return an context from the initial call to `initialize`? Is it happy if it's passed a 0-byte input buffer?

src/tls_stream.rs
```diff
@@ -193,62 +231,57 @@ impl<S> TlsStream<S>
...
+                                                    0,
+                                                    0,
+                                                    &mut inbuf_desc,
+                                                    0,
+                                                    ptr::null_mut(),
+                                                    &mut outbuf_desc,
+                                                    &mut attributes,
+                                                    ptr::null_mut())
+            };
+            debug!("attributes: {:x}", attributes);
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -258,6 +291,21 @@ impl<S> TlsStream<S>
...
+                    // In other words, if we were to set `accept_first` to
+                    // `false` after the literal first call to
+                    // `AcceptSecurityContext` while the call returned
+                    // INCOMPLETE_MESSAGE, the next call would return an error.
+                    //
+                    // For that reason we only set `accept_first` to false here
+                    // once we've actually successfully received the full
+                    // "token" from the client.
+                    self.accept_first = false;
+                    debug!("continue needed");
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -269,8 +317,12 @@ impl<S> TlsStream<S>
                     self.needs_read = self.enc_in.position() == 0;
                     self.out_buf.get_mut().extend_from_slice(&to_write);
                 }
-                winapi::SEC_E_INCOMPLETE_MESSAGE => self.needs_read = true,
+                winapi::SEC_E_INCOMPLETE_MESSAGE => {
+                    debug!("incomplete message");
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -269,8 +317,12 @@ impl<S> TlsStream<S>
...
+                    self.needs_read = true;
+                }
                 winapi::SEC_E_OK => {
+                    debug!("security context ok");
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -294,7 +346,11 @@ impl<S> TlsStream<S>
                         *more_calls = false;
                     }
                 }
-                _ => return Err(io::Error::from_raw_os_error(status as i32)),
+                _ => {
+                    let err = io::Error::from_raw_os_error(status as i32);
+                    debug!("initialize/accept error: {:?}", err);
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -345,6 +401,12 @@ impl<S> TlsStream<S>
     }
 
     fn validate(&mut self) -> io::Result<()> {
+        // If we're accepting connections then we don't perform any validation
+        // fo the remote certificate, that's what they're doing!
```
sfackler: Will this continue being true when we add support for client certificate auth?

src/tls_stream.rs
```diff
@@ -381,7 +443,9 @@ impl<S> TlsStream<S>
             if res == winapi::TRUE {
                 CertChainContext(cert_chain as *mut _)
             } else {
-                return Err(io::Error::last_os_error());
+                let err = io::Error::last_os_error();
+                debug!("CertGetCertificateChain error -- {:?}", err);
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -406,11 +470,15 @@ impl<S> TlsStream<S>
                                                                 &mut para,
                                                                 &mut status);
             if res == winapi::FALSE {
-                return Err(io::Error::last_os_error());
+                let err = io::Error::last_os_error();
+                debug!("CertVerifyCertificateChainPolicy error {:?}", err);
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -406,11 +470,15 @@ impl<S> TlsStream<S>
...
+                return Err(err)
             }
 
             if status.dwError != winapi::ERROR_SUCCESS {
-                return Err(io::Error::from_raw_os_error(status.dwError as i32));
+                let err = io::Error::from_raw_os_error(status.dwError as i32);
+                debug!("stat.dwError error {:?}", err);
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -449,44 +517,21 @@ impl<S> TlsStream<S>
         let count = size - nread;
 
         if count > 0 {
-            unsafe {
-                let src = &self.enc_in.get_ref()[nread] as *const _;
-                let dst = self.enc_in.get_mut().as_mut_ptr();
-    
-                ptr::copy(src, dst, count);
-            }
+            self.enc_in.get_mut().drain(..nread);
```
sfackler: 💯

src/tls_stream.rs
```diff
@@ -531,7 +576,11 @@ impl<S> TlsStream<S>
                     self.needs_read = self.enc_in.position() == 0;
                     Ok(())
                 }
-                e => Err(io::Error::from_raw_os_error(e as i32)),
+                e => {
+                    let err = io::Error::from_raw_os_error(e as i32);
+                    debug!("decrypt error {:?}", e);
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -588,12 +622,71 @@ impl<S> TlsStream<S>
                     self.out_buf.set_position(0);
                     Ok(())
                 }
-                err => Err(io::Error::from_raw_os_error(err as i32)),
+                err => {
+                    let err = io::Error::from_raw_os_error(err as i32);
+                    debug!("failed to encrypt: {:?}", err);
```
sfackler: ^

src/tls_stream.rs
```diff
@@ -588,12 +622,71 @@ impl<S> TlsStream<S>
...
+            }
+            Err(e) => {
+                debug!("handshake error: {:?}", e);
+                Err(HandshakeError::Failure(e))
+            }
+        }
+    }
+}
+
+impl<S: fmt::Debug + Any> Error for HandshakeError<S> {
```
sfackler: Ah, nevermind, the Error comment above.

Can these impls be moved next to the type definition though?

src/tls_stream.rs
```diff
@@ -588,12 +622,71 @@ impl<S> TlsStream<S>
...
+        match *self {
+            HandshakeError::Failure(ref e) => Some(e),
+            HandshakeError::Interrupted(_) => None,
+        }
+    }
+}
+
+impl<S> fmt::Display for HandshakeError<S> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let msg = match *self {
```
sfackler: This can just delegate to description.

sfackler: ... but should then also print the cause if present.

test/localhost.crt
```diff
@@ -0,0 +1,31 @@
+-----BEGIN CERTIFICATE-----
```
sfackler: If everyone's going to be making their own copy of these they probably shouldn't be tracked, right?

sfackler:
So to make sure I understand it, the plan for trust roots is for each person to generate a CA, have it sign a certificate, delete the CA private key, install the CA certificate in the user's trust store, and then use the certificate in tests? That seems okay, though i'm a bit worried about adding to the users's trust root, even if we are disposing of the private key.

A possible alternative is to have each test create an ephemeral CA with an expiry of like 1 minute, install it and then uninstall when the test's done.

src/cert_store.rs
```diff
@@ -63,11 +70,95 @@ impl CertStore {
...
+            if res.is_null() {
+                Err(io::Error::last_os_error())
+            } else {
+                Ok(CertStore(res))
+            }
+        }
+    }
+
+    /// Returns an iterator over the certificates in this certificate store.
+    pub fn iter(&mut self) -> CertStoreIter {
```
sfackler: Should this be called `certs` or something like that? Certificate stores also store CTLs etc.

sfackler:
I should note that I *don't* think it's okay to have a git-tracked CA that people are supposed to install.

steffengy:
Also for creating the certificates it might be advisable to just use powershell?
Is there any reason, we're not simply doing:
```ps
$cert = New-SelfSignedCertificate -DnsName localhost -CertStoreLocation cert:Localmachine\My
Export-Certificate -Cert $cert -FilePath localhost.cer -Type CERT
$pwd = ConvertTo-SecureString -String "foobar" -Force –AsPlainText
Export-PfxCertificate -Cert $cert -FilePath localhost.pfx -Password $pwd
Import-PfxCertificate -FilePath localhost.pfx -CertStoreLocation cert:\LocalMachine\Root -Password $pwd
```
in combination with [these changes (GIST)] (https://gist.github.com/steffengy/e284f591fe26a3e6a0bab7c65999b46b) to the PR?

(That should ofcourse also work for the user's personal store, for appveyor we'll still need the machine store)

src/cert_store.rs
alexcrichton: Maybe it could just be called `CertStore` and dropped entirely? I'm not sure if there are restrictions on calling the two methods here on arbitrary `CertStore` types, but the windows apis at least allow it!

src/cert_context.rs
alexcrichton: I'll rename to `sign_hash_algorithms` for now, but I also don't mind dropping it. I thought I was going to need it for tests but ended up not needing it.

src/ctl_context.rs
alexcrichton: Oh I was just using this for debugging, would you prefer if all debugging statements were removed? I figured I'd leave them in to see if they're helpful at some later date.

src/schannel_cred.rs
alexcrichton: Oh this is kinda what we were talking about at the meetup, where the consumption method is `mut` to allow transferring ownership of any resource that needs to be consumed. Right now it's not taken advantage of but it's a bit more future-proof to adding more flavorful configuration perhaps. (not sure if it'll ever be needed in this case.

src/test.rs
alexcrichton: Sure! I'm not actually aware of one (not that I've done a lot of digging), but do you have one in mind?

src/tls_stream.rs
alexcrichton: Ok, I'll stick with `SchannelCred` as you already specify inbound/outbound

src/tls_stream.rs
alexcrichton: Weird I thought I added that

alexcrichton: Ah wait yeah, it's just below

src/tls_stream.rs
alexcrichton: Yeah unfortunately returning a context in the initial call to `initialize` is basically guaranteed to fail because there's no input data. So we'd always have to call it again and throw away the context from the first call (I don't think there is one, though).

src/tls_stream.rs
alexcrichton: Probably not, but I'm not sure how that works at all :(

src/tls_stream.rs
alexcrichton: Yeah I was wondering the best way to do this, the `Error` impl requires the `Any` bound but the `Display` here impl doesn't, so I figured I'd make this one more general. I also think it's somewhat best practice to not print information about the cause because that's what the renderer for the entire chain of error objects should in theory be responsible for I think.

test/localhost.crt
alexcrichton: Oh I was thinking that everyone would just use these that are checked in. The script is just there in case anyone wants to run it again in the future (e.g. tweak the various parameters here and there) or create new certs for new tests.

src/cert_store.rs
alexcrichton: Seems reasonable to me!

alexcrichton:
@sfackler yeah it was actually my intention to check in all the certificates to git and contributors who wanted to run these tests would add the certificate to their trusted root store. I figured it's not that much of a risk as the private key is deleted immediately for the CA, so it can only ever sign one key (and that's just for localhost).

That being said I wouldn't mind creating ephemeral keys, I just couldn't figure out how to put something into the root trust store. I wouldn't mind trying it out though, means we may not need to check in any keys at all! You wouldn't happen to know the API off-hand to do that though would you?

@steffengy yeah I'd be fine with that although I'd prefer to make API calls directly rather than user power shell itself (just so a vanilla `cargo test` works), but it'd all boil down to the same thing I think. Do you not need elevated permissions or dialog clicks to add something to the system certificate store?

alexcrichton:
@sfackler I've left the debug statements in for now but if you'd prefer I remove them all just wanted to confirm and I'll do so.

steffengy:
@alexcrichton 
You need elevated permissions (UAC) for the system certificate store and somehow a dialog pops up for the user (Root-) certificate store, as we've seen with the appveyor tests.

Using the API directly should show the same behavior, which might make a vanilla ```cargo test``` difficult to implement that way (Except if you rely on elevated permissions). I'll do some investigating.

steffengy:
@alexcrichton 
https://gist.github.com/steffengy/fcbf37eda64d1f7ce982b3ffcbf97bb9

With these changes (requires some cleaning up), a vanilla `cargo test` works.
It works the following way:
- If **no certificate exists in the test folder**, it creates a self-signed one and
stores it there.
- If **a certificate exists in the test folder**, it searches for it in the **Root LocalMachine**
store,   
checks the validity (if it has expired yet) and **either** 
   - **uses the stored one** (if found and valid)
   - or **creates a new one**, as if none exists.

The certificate is valid for 5 minutes, which should (currently) be enough to run all tests.

src/cert_store.rs
sfackler: I had them separate to provide a bit more type safety - you can't add certificates to a collection cert store, or add other cert stores to an in memory store for example.

src/ctl_context.rs
sfackler: Yeah ideally I'd remove them all.

src/tls_stream.rs
sfackler: I have never seen anything print the entire error chain :P

This is a problem in itself, but in the meantime the convention seems to be something like this:

```rust
fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
    try!(fmt.write_str(self.description));
    if let Some(e) = self.cause() {
       try!(write!(fmt, ": {}", e));
    }
    Ok(())
}
```

sfackler:
> I figured it's not that much of a risk as the private key is deleted immediately for the CA, so it can only ever sign one key (and that's just for localhost).

Whatever random person last regenerated the CA *claims* that they deleted the private key.

If elevated permissions are required to do the ephemeral thing, then a long-lived per-developer CA seems fine for now. I'd ideally like to get rid of the openssl stuff (dogfood!) but that's not super necessary.

src/schannel_cred.rs
sfackler: Looking at the docs, the only thing that might apply here is that you can provide a certificate store with trusted roots for client certificates, but I'm not sure about the ownership semantics of that.

alexcrichton: Would you prefer this stays as `&self` (I just want to get this landed)

sfackler: Yeah let's leave it like that for now if possible.

alexcrichton:
Thanks for the commit @steffengy! I'll take that strategy and wrap it all up in some APIs as well.

steffengy:
@alexcrichton 
Since we cannot bypass the validation, the short validity may be counter-productive.

Probably setting the validity to **one year** is the best option, should not be
a high security risk, since it's all locally generated.
962f7f928beefeb6fb6cf48d6e7d58bebd0498f0 should do that.

This would only require elevation on the first run, and probably is indeed the best option.
(@sfackler also seems to prefer this)

alexcrichton:
Ok, I've pushed an updated using @steffengy's strategy. I opted to have certificates valid for a day as it basically just means that once a day if you run tests you have to click a dialog twice (once to delete, once to add), and it's somewhat less dangerous than a year.

I added an API or two along the way, but we still deal with creating the self-signed certificate in a raw fashion. Outstanding questions are:

* Do we want to just remove `CertStore::memory`? I don't think the final version is going to end up using it.
* Do we want to just not expose `CtlContext` for now?

README.md
```diff
@@ -6,3 +6,29 @@ schannel-rs [![Build status](https://ci.appveyor.com/api/projects/status/vefyaua
...
+### Running tests
+
+Most tests can be executed with a `cargo test`, but some tess require SSL
+certificates to be trusted in the root store on Windows. For this reason, a
+vanilla `cargo test` on a fresh checkout will likely fail tests. There are two
+options here:
+
+1. Export the `SCHANNEL_RS_SKIP_SERVER_TESTS=1` environment variable, forcing
+   the test suite to skip all related tests.
+2. Add the `test/schannel-ca.crt` certificate to the root trust store on
```
steffengy: minor: This section doesn't seem quite up to date anymore.

sfackler:
Let's leave `CtlContext` private until we have a chance to add in signing, yeah, and same with `Memory`.

alexcrichton:
Ok, left both of those private and I just removed the bits in the README

src/cert_context.rs
```diff
@@ -1,13 +1,206 @@
...
+    /// current time.
+    pub fn is_time_valid(&self) -> io::Result<bool> {
+        let ret = unsafe {
+            crypt32::CertVerifyTimeValidity(ptr::null_mut(), (*self.0).pCertInfo)
+        };
+        Ok(ret == 0)
+    }
+
+    /// Deletes this certificate from its certificate store.
+    pub fn delete(self) -> io::Result<()> {
```
sfackler: Is the certificate not valid after `CertDeleteCertificateFromStore` is called?

sfackler: Nevermind, read the docs.

src/cert_store.rs
```diff
@@ -63,11 +72,145 @@ impl CertStore {
...
+                Ok(CertStore(store))
+            }
+        }
+    }
+
+    /// Imports a PKCS#12-encoded key/certificate pair, returned as a
+    /// `CertStore` instance.
+    ///
+    /// The password must also be provided to decrypt the encoded data.
+    pub fn import_pkcs12(data: &[u8], password: &str) -> io::Result<CertStore> {
```
sfackler: A password is always required?

src/cert_context.rs
alexcrichton: The ~magic~ of ownershpi

src/cert_store.rs
```diff
@@ -63,11 +72,145 @@ impl CertStore {
...
+
+    /// Adds a certificate context to this store.
+    ///
+    /// This function will add the certificate specified in `cx` to this store.
+    /// A copy of the added certificate is returned.
+    pub fn add_cert(&mut self, cx: &CertContext) -> io::Result<CertContext> {
+        unsafe {
+            let flags = winapi::CERT_STORE_ADD_REPLACE_EXISTING;
+            let mut ret = ptr::null();
+            let res = crypt32::CertAddCertificateContextToStore(self.0,
```
sfackler: I think we'd probably want to expose this a bit more powerfully (e.g. the different replacement strategies)

src/cert_store.rs
alexcrichton: Looks like the API documents NULL as valid. I couldn't figure out how to create a PKCS#12-encoded thing without it but I'll make this `Option`.

steffengy: I wouldn't know that i's possible with the windows crypto-api 
(atleast it doesn't seem to be used very often, PowerShell requires a password always).
It's possible with openssl though.

alexcrichton:
Made the password on pkcs12 optional and added an enum argument to `add_cert`

src/cert_store.rs
```diff
@@ -63,11 +127,152 @@ impl CertStore {
...
+
+    /// Adds a certificate context to this store.
+    ///
+    /// This function will add the certificate specified in `cx` to this store.
+    /// A copy of the added certificate is returned.
+    pub fn add_cert(&mut self,
+                    cx: &CertContext,
+                    how: CertAdd) -> io::Result<CertContext> {
+        unsafe {
+            let how = how as winapi::DWORD;
```
sfackler: Huh, didn't know you could downcast an enum without an explicit repr.

sfackler:
Woo!

PR #9, opened 2016-08-01
Read approximation on SEC_E_INCOMPLETE_MESSAGE

steffengy:
Basic implementation to use the approximation returned by the crypto api,
to reduce the amount of API calls.

This essentially reads atleast the approximated amount of bytes.

### DIFF
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -139,7 +139,7 @@ impl Builder {
                 more_calls: true,
                 shutting_down: false,
             },
-            needs_read: true,
+            needs_read: 1,
             dec_in: Cursor::new(Vec::new()),
             enc_in: Cursor::new(Vec::new()),
             out_buf: Cursor::new(buf.map(|b| b.to_owned()).unwrap_or(Vec::new())),
@@ -171,7 +171,7 @@ pub struct TlsStream<S> {
     state: State,
     accept: bool,
     accept_first: bool,
-    needs_read: bool,
+    needs_read: usize,
     // valid from position() to len()
     dec_in: Cursor<Vec<u8>>,
     // valid from 0 to position()
@@ -280,7 +280,7 @@ impl<S> TlsStream<S>
                     more_calls: true,
                     shutting_down: true,
                 };
-                self.needs_read = false;
+                self.needs_read = 0;
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/tls_stream.rs
```diff
@@ -541,17 +545,27 @@ impl<S> TlsStream<S>
...
+            let min_len = cmp::max(1024, self.needs_read + 2 * existing_len);
+            if self.enc_in.get_ref().len() < min_len {
+                self.enc_in.get_mut().resize(min_len, 0);
+            }
+            let nread = {
+                let buf = &mut self.enc_in.get_mut()[existing_len..];
+                try!(self.stream.read(buf))
+            };
+            self.enc_in.set_position((existing_len + nread) as u64);
+            self.needs_read = self.needs_read.checked_sub(nread).unwrap_or(0);
```
sfackler: This can just be saturating_sub(nread)

src/tls_stream.rs
```diff
@@ -595,19 +609,25 @@ impl<S> TlsStream<S>
...
+                        bufs[1].cbBuffer as usize
+                    } else {
+                        1
+                    };
                     Ok(())
                 }
-                state @ winapi::SEC_I_CONTEXT_EXPIRED |
-                state @ winapi::SEC_I_RENEGOTIATE => {
+                winapi::SEC_I_CONTEXT_EXPIRED => {
+                    self.shutdown()
```
sfackler: This is the fix for that wikipedia issue you ran into, right?

src/tls_stream.rs
```diff
@@ -541,17 +545,27 @@ impl<S> TlsStream<S>
...
     fn read_in(&mut self) -> io::Result<usize> {
-        let existing_len = self.enc_in.position() as usize;
-        let min_len = cmp::max(1024, 2 * existing_len);
-        if self.enc_in.get_ref().len() < min_len {
-            self.enc_in.get_mut().resize(min_len, 0);
+        let mut sum_nread = 0;
+
+        while self.needs_read > 0 {
+            let existing_len = self.enc_in.position() as usize;
+            let min_len = cmp::max(1024, self.needs_read + 2 * existing_len);
```
sfackler: This seems a little suspicious - maybe it should be something like `max(max(1024, existing_len * 2), self.needs_read)`? We want to make sure our buffer is larger than 1023, larger than the amount we want to read, and is doubling in size past that point.

sfackler:
Looks good to me other than a couple nits!

src/tls_stream.rs
steffengy: not wikipedia only, but yeah also.

src/tls_stream.rs
steffengy: Seems better, although it looks a bit strange. I'll add it

PR #10, opened 2016-08-16
Support PKCS #12 imports

sfackler:
Also some small cleanup here and there.


src/cert_store.rs
```diff
@@ -263,6 +268,66 @@ impl<'a> Iterator for Certs<'a> {
...
+    /// If set, all extended properties of the certificate will be imported.
+    pub fn include_extended_properties(&mut self,
+                                       include_extended_properties: bool)
+                                       -> &mut PfxImportOptions {
+        self.flag(PKCS12_INCLUDE_EXTENDED_PROPERTIES, include_extended_properties)
+    }
+
+    fn flag(&mut self, flag: winapi::DWORD, set: bool) -> &mut PfxImportOptions {
+        if set {
+        self.flags |= flag;
```
steffengy: @sfackler 
Is the indentation intended here and below intended? Looks a bit strange to me.

sfackler: Oops, fixed.

PR #15, opened 2016-12-15
Add CertStore::export_pkcs12

alexcrichton:
This exposes the ability to export a `CertStore` as a PKCS#12 archive. Along
the way this also exports the `memory` constructor to allow easily converting a
`CertContext` with a private key to a PKCS#12 archive. Along the way this
removes the `Memory` type and moves the functions to the `CertStore` type
directly.

### DIFF
--- a/src/cert_store.rs
+++ b/src/cert_store.rs
@@ -115,24 +115,6 @@ pub enum CertAdd {
 }
 
 impl CertStore {
-    /// Creates a new in-memory certificate store which certificates and CTLs
-    /// can be added to.
-    #[allow(dead_code)]
-    /* pub */ fn memory() -> io::Result<Memory> {
-        unsafe {
-            let store = crypt32::CertOpenStore(winapi::CERT_STORE_PROV_MEMORY as winapi::LPCSTR,
-                                               0,
-                                               0,
-                                               0,
-                                               ptr::null_mut());
-            if store.is_null() {
-                Err(io::Error::last_os_error())
-            } else {
-                Ok(Memory(CertStore(store)))
-            }
-        }
-    }
-
     /// Opens up the specified key store within the context of the current user.
     ///
     /// Known valid values for `which` are "Root" and "My".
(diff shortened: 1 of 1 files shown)


alexcrichton:
The rationale behind this that during local testing when using native-tls I want to create a `CertContext` with winapi functions but then convert that to a `TlsAcceptor` which involves going through the middle-man of a PKCS#12 archive. Feedback welcome though!

steffengy:
from my point of view +1 since supporting basic certificate functionality makes sense.
(required for tests or basic use)

(Yes it would be better to have these sort of APIs in a seperate crate but until it's a bit
more "fleshed out" an extraction into a separate crate doesn't make sense yet,
from my perspective - and one can pretty easily move it to one at a a later stage)

Unifying memory and CertStore is in my oppinion also ok, since I think
it was an abstraction, which didn't really have a use (atleast none i know of)?

@sfackler any objections or feedback from your side?

sfackler:
CertStores are reference counted and I think I was originally unsure if those mutation operations are appropriately thread safe. If they are then this seems fine.

steffengy:
Oh, so it was to prevent people from sharing the underlying CertStore across threads, to prevent
concurrent modifications?

That indeed will be an issue while sharing them is allowed.

There is no documentation regarding thread safety on any of the mutations above,
so I think the sane choice is to assume they aren't.

The [.NET documentation for x509Store](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509store(v=vs.110).aspx) also hints: 
"Any instance members are not guaranteed to be thread safe."

Technically I also assume they don't take the overhead of ensuring a consistent view
of a certificate store by doing locking or anything similar.

So it probably makes sense to expose a certificate store you can mutate but not share across
threads like before?

Still not sure how that'd work with some "physical certstore" like the current user's cert store,
if it'd be mutated from multiple threads.

alexcrichton:
Good points! Perhaps I could just remove the `Sync` and `Send` from `CertStore`? I think that'd address those points, right?

steffengy:
That'd address those points yes.

In terms of consistency I'm unsure if it's the best solution though,
since some other wrapping types (`CertContext ` and `CtlContext`) 
also implement `Sync` and `Send`.

It also might lead to usability implications, which might've been @sfackler's 
original rationale for making those types `Sync` and `Send`?

In my opinion it still should be cleaner to share a "new" handle to that 
store/context/whatever, when you need access from multiple threads? 
Then it would make sense to remove `Sync` and `Send` on those types.

Still not quite sure how significant the disadvantages in each case would be.

sfackler:
`CertStore` has to be `Sync` and `Send` - otherwise `TlsStream` wouldn't be!

Having a separate builder type and keeping `CertStore` immutable avoids these issues.

alexcrichton:
I'm a little lost now, unfortunately. Could you summarize the changes you'd like to see?

sfackler:
Keep `CertStore` immutable and add back the `Memory` type.

alexcrichton:
And to clarify, `CertStore::add_cert` already exists, but that's ok? Should change?

alexcrichton:
@sfackler thought on [this](https://github.com/steffengy/schannel-rs/pull/15#issuecomment-267796502)? Should those methods be deprecated now if we're worried about reference counting?

In any case I've pushed those changes

sfackler:
Ok, so I think we should deprecate `CertStore::add_cert`, add a `CertStoreBuilder` type that has `memory`, `open_local_machine` and `open_current_user` constructors and does have that method.

alexcrichton:
Perhaps that could happen in a separate PR? I'm already running a little thin as-is and all I wanted to do was update tokio-tls...

I don't really understand the reference counting issues here, and it sounds like they need to be thought through rather than bolting on a solution here?

sfackler:
Works for me.

PR #17, opened 2017-02-27
support cross-checking the root certificate against the provided store

steffengy:
This trusts self-signed certificates, if they are in the provided certificate store.

The implementation as of now isn't quite nice yet,
but it's the first draft that seems to do what's wanted,
so at the moment early-review, until I get to 
polish some parts of it.


sfackler:
Nice! I'll take a look tonight!

src/cert_context.rs
```diff
@@ -23,6 +24,20 @@ pub struct CertContext(winapi::PCCERT_CONTEXT);
 unsafe impl Sync for CertContext {}
 unsafe impl Send for CertContext {}
 
+// TODO: unsure if this is a solid decision instead of a fn `matches`
+// should work anyways
+impl cmp::PartialEq for CertContext {
+    fn eq(&self, other: &CertContext) -> bool {
+        let ret = unsafe {
+            crypt32::CertComparePublicKeyInfo(
```
sfackler: @bascule what's the most reasonable way to compare two certificates for equality? Looks like OSCrypto is comparing cert hashes: https://github.com/wbond/oscrypto/blob/master/oscrypto/_win/tls.py#L608

src/cert_context.rs
```diff
@@ -204,6 +219,14 @@ impl CertContext {
             }
         }
     }
+
+    fn get_public_key_info(&self) -> winapi::PCERT_PUBLIC_KEY_INFO {
+        unsafe {
+            let cert_info = *(*self.0).pCertInfo;
+            // TODO: since we do not expect any mutation this should be safe?
```
sfackler: Yeah this is fine. Const-correctness in C APIs is dubious at best usually.

src/tls_stream.rs
```diff
@@ -501,6 +502,30 @@ impl<S> TlsStream<S>
...
+                if cert_chain.cChain > 0 {
+                    let first_rgp_chain = **cert_chain.rgpChain;
+                    if first_rgp_chain.cElement > 0 {
+                        let elements = slice::from_raw_parts(
+                            first_rgp_chain.rgpElement as *mut &mut winapi::CERT_CHAIN_ELEMENT,
+                            first_rgp_chain.cElement as usize);
+                        let final_element = elements.last().unwrap();
+                        
+                        let root_cert = CertContext::from_inner(final_element.pCertContext);
+                        for cert in store.certs() {
```
sfackler: This might be a bit simpler as `store.certs().any(|c| c == root_cert)`.

src/tls_stream.rs
```diff
@@ -501,6 +502,30 @@ impl<S> TlsStream<S>
         };
 
         unsafe {
+            // check if we trust the root-CA explicitly
+            let mut para_flags = winapi::CERT_CHAIN_POLICY_IGNORE_ALL_REV_UNKNOWN_FLAGS;
+            if let Some(ref mut store) = self.cert_store {
+                let cert_chain = *cert_chain.0;
+                if cert_chain.cChain > 0 {
+                    let first_rgp_chain = **cert_chain.rgpChain;
```
sfackler: I think we want to be looking at the final chain according to the docs: https://msdn.microsoft.com/en-us/library/windows/desktop/aa377182(v=vs.85).aspx (see the rgpChain section).

src/tls_stream.rs
```diff
@@ -501,6 +502,30 @@ impl<S> TlsStream<S>
...
+                    if first_rgp_chain.cElement > 0 {
+                        let elements = slice::from_raw_parts(
```
sfackler: It's probably simpler to just do `first_rgp_chain.rgpElement.offset(first_rgp_chain.cElement as isize)`.

sfackler:
The docs here will need to be updated: http://steffengy.github.io/schannel-rs/doc/schannel/tls_stream/struct.Builder.html#method.cert_store

src/cert_context.rs
steffengy: @sfackler 
Yeah I first thought about simply using `.sha1` but after SHAttered I prefered SHA256, which isn't easily accessible yet, so I settled on just comparing the public key for now.
I'll compare SHA256 hashes in the next draft.

src/tls_stream.rs
steffengy: seems to be exactly what OSCrypto and Microsoft is doing, doesn't it?
https://github.com/Microsoft/Windows-classic-samples/blob/master/Samples/Win7Samples/security/cryptoapi/VerifyNameTrust/VerifyNameTrust/VerifyNameTrust.cpp#L172

src/tls_stream.rs
steffengy: Did that in the first draft, personally I prefer this one, since there's less dereferencing.
and you'd need to do ```first_rpg_chain.cElement as isize - 1```

I think the current one is easier to get correct, so I'd prefer to keep it this way
(mainly because it just subjectively looks "nicer")?

src/tls_stream.rs
sfackler: Oh right, I misread.

src/tls_stream.rs
```diff
@@ -501,6 +501,30 @@ impl<S> TlsStream<S>
...
+                            first_rgp_chain.rgpElement as *mut &mut winapi::CERT_CHAIN_ELEMENT,
+                            first_rgp_chain.cElement as usize);
+                        let final_element = elements.last().unwrap();
+                        
+                        let root_cert = CertContext::from_inner(final_element.pCertContext);
+                        for cert in store.certs() {
+                            if try!(cert.is_equal(&root_cert)) {
```
sfackler: We'll double free if this returns early.

steffengy: Ouch.
The current revision contains an implementation, which doesn't
need 2 mem::forget's and might be the most elegant solution 
(without using a RAII-"FreeGuard", that requires more boilerplate).

A simple `match` would require 2 mem::forget's, so I slightly prefer this one.
Are there any disadvantages I didn't consider there? 
(except maybe a tiny performance impact?)

steffengy:
@sfackler 
any other/unresolved issues with this?

src/tls_stream.rs
```diff
@@ -501,6 +501,34 @@ impl<S> TlsStream<S>
...
+                        let elements = slice::from_raw_parts(
+                            first_rgp_chain.rgpElement as *mut &mut winapi::CERT_CHAIN_ELEMENT,
+                            first_rgp_chain.cElement as usize);
+                        let final_element = elements.last().unwrap();
+                        
+                        let root_cert = CertContext::from_inner(final_element.pCertContext);
+                        // find the first cert that matches this root_cert
+                        let cert_matches_result = store.certs()
+                             .map(|cert| cert.is_equal(&root_cert))
+                             .filter(|result| result.is_err() || *result.as_ref().ok().unwrap())
```
sfackler: Can we avoid this unwrap?

src/cert_context.rs
```diff
@@ -130,6 +130,13 @@ impl CertContext {
         }
     }
 
+    /// Compares two certificates and returns true, if they are identical
+    pub fn is_equal(&self, other: &CertContext) -> io::Result<bool> {
```
sfackler: If all we're doing is comparing for equality we should be able to just compare the encoded bytes directly to avoid the Err case here.

sfackler:
Sorry, was super busy for the last couple of days!

src/cert_context.rs
steffengy: good point, luckily that's available through CERT_CONTEXT. updated the PR.

steffengy:
(nightly) tests fail until https://github.com/rust-lang/rust/pull/40450 lands.
(Cargo Version print crashes in current nightlies)

src/cert_context.rs
```diff
@@ -204,6 +212,32 @@ impl CertContext {
             }
         }
     }
+
+    fn get_sha256_hash(&self) -> io::Result<[u8; 32]> {
```
sfackler: This is dead code now.

sfackler:
Looks good other than getting rid of a bit of dead code!

PR #18, opened 2017-03-13
Support PEM-formatted certificates

sfackler:


src/cert_context.rs
```diff
@@ -44,6 +43,32 @@ impl CertContext {
         }
     }
 
+    /// Decodes a PEM-formatted X509 certificate.
+    pub fn from_pem(pem: &str) -> io::Result<CertContext> {
+        unsafe {
+            assert!(pem.len() <= winapi::DWORD::max_value() as usize);
+
+            // the docs claim that you can pass null as the out pointer and
```
steffengy: very strange, I'll look into it, but fortunately quite a low-priority issue.

steffengy: I cannot reproduce that:
https://github.com/steffengy/schannel-rs/commit/12e010b0917df698fb3479c1bf1f5df26f5532dc
works fine locally.
Also AppVeyor tests passed:
https://ci.appveyor.com/project/steffengy/schannel-rs/build/1.0.104

Can you confirm that this commit doesn't work for you locally?

sfackler:
I saw reads or writes to address 0, but if it works for you then I must have just messed something up :)

I'll make sure that works locally for me when I get home and then switch over.

sfackler:
Looks like it's working for me now as well - not sure what I was doing before...

PR #21, opened 2017-03-17
Don't automatically shutdown on receipt of a close_notify

### LINKED ISSUE #19: Should errors in shutdown be ignored?
Over at tokio-tls we've got a [smoke test](https://github.com/tokio-rs/tokio-tls/blob/44a5c9edab1a53e5bec9b1ae00a5ec919b108f22/tests/smoke.rs#L616-L649) which just tests a simple TLS handshake and sending some bytes from a client to a server, but notably limits I/O reads/writes to only progress one byte at a time. This is intended to stress retry logic and ensure that at least most paths are covered.

Previously this test passed, but recently a change was made to properly issue a shutdown operation from the server side when it's done writing (the test just writes a bunch of bytes from the server to the client). This unfortunately has started making the test fail with schannel!

It looks like schannel is acknowledging a shutdown and [attempting to run its own](https://github.com/steffengy/schannel-rs/blob/6899374578f8de72dae72fcdd0cbfa304ac856cd/src/tls_stream.rs#L648-L650), but then during that function a WSAECONNRESET is received during a TCP write (b/c the other end has dropped at that point).

I'm curious, have you seen this before? Are you familiar with behavior like this? I'm not sure if this is a bug in the test (e.g. requiring a call here or there) or if it's a case where the error should be ignored. The test passes on OSX/Linux FWIW, but Windows (schannel) may just be exposing a surprising edge case.

In any case, thoughts would be welcome!

sfackler:
Closes #19

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -416,6 +416,7 @@ fn accept_a_socket() {
         stream.write_all(&[1, 2, 3, 4]).unwrap();
         stream.flush().unwrap();
         assert_eq!(stream.read(&mut [0; 1024]).unwrap(), 4);
+        stream.shutdown().unwrap();
     });
 
     let stream = listener.accept().unwrap().0;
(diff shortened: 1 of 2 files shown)


sfackler:
@alexcrichton

alexcrichton:
👍 works for me!

alexcrichton:
Thanks!

sfackler:
Cool - @steffengy does this look reasonable to you as well?

steffengy:
@sfackler fine with me

PR #24, opened 2017-05-04
Pass ISC_REQ_USE_SUPPLIED_CREDS to InitializeSecurityContext

### LINKED ISSUE #23: Can't connect to cert-requesting servers without client cert
When using this crate to connect to a TLS server that requests a client certificate, and no client certificate was configured by the program, the connection fails with the error `SEC_I_INCOMPLETE_CREDENTIALS`.

```rust
extern crate schannel;

use std::net::TcpStream;
use std::io::{self, Write};

use schannel::schannel_cred::{SchannelCred, Direction};
use schannel::tls_stream::{Builder as TlsStreamBuilder};

fn main() {
	let s = TcpStream::connect(std::env::args().nth(1).unwrap()).unwrap();
    let mut s = TlsStreamBuilder::new().connect(SchannelCred::builder().acquire(Direction::Outbound).unwrap(), s).unwrap();
	s.write_all(b"GET / HTTP/1.0\r\n\r\n").unwrap();
	io::copy(&mut s, &mut io::stdout()).unwrap();
}
```

```
$ cargo run auth.mit.edu:444
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Failure(Error { repr: Os { code: 590624, message: "The credentials supplied were not complete, and could not be verified. Additional information can be returned from the context." } })', C:\bot\slave\stable-dist-rustc-win-msvc-64\build\src\libcore\result.rs:868
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

I'm actively investigating solutions.

jethrogb:
Fixes #23. As mentioned in there, not sure if this is the behavior you want.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,7 +36,7 @@ const ACCEPT_REQUESTS: winapi::c_ulong =
 const INIT_REQUESTS: winapi::c_ulong =
     winapi::ISC_REQ_CONFIDENTIALITY | winapi::ISC_REQ_INTEGRITY | winapi::ISC_REQ_REPLAY_DETECT |
     winapi::ISC_REQ_SEQUENCE_DETECT | winapi::ISC_REQ_MANUAL_CRED_VALIDATION |
-    winapi::ISC_REQ_ALLOCATE_MEMORY | winapi::ISC_REQ_STREAM;
+    winapi::ISC_REQ_ALLOCATE_MEMORY | winapi::ISC_REQ_STREAM | winapi::ISC_REQ_USE_SUPPLIED_CREDS;
 
 trait Inner<T> {
     unsafe fn from_inner(t: T) -> Self;


sfackler:
From the documentation, it sounds like this flag may not solve the problem described in #23?

> When this flag is specified, Schannel will return SEC_I_INCOMPLETE_CREDENTIALS to the client when the server requests authentication and the client has not previously supplied a certificate.

https://msdn.microsoft.com/en-us/library/windows/desktop/aa378819(v=vs.85).aspx

steffengy:
@jethrogb's testing and common usage such as 
https://github.com/curl/curl/blob/07fd7871b38cc8472c3806e254ba4062e3adeae0/lib/vtls/schannel.c#L594

seems to indicate that the documentation is off (once again).

sfackler:
We would need a similar check for that error code, right: https://github.com/curl/curl/blob/07fd7871b38cc8472c3806e254ba4062e3adeae0/lib/vtls/schannel.c#L592

steffengy:
In short, If we want the same behavior, yes.

#### Behavior 1 CURL
- First attempt to connect the server, leaving users certificates to schannel (which will look through several stores trying to find something that might work and sending it to the server)
- If the connection fails with `SEC_I_INCOMPLETE_CREDENTIALS`, pass `ISC_REQ_USE_SUPPLIED_CREDS` to prevent SCHANNEL from sending any certificates (since it/we do not pass any to it). 

==> Potentially unwanted that certificates are sent automatically?

#### Other behavior (as of this PR)
The more sensible alternative would be to rely on `ISC_REQ_USE_SUPPLIED_CRED` for all cases.
This would [apparently](https://groups.google.com/forum/#!msg/microsoft.public.platformsdk.security/lb-9guU8-D8/tgBBECWKyLYJ) only send the certificate associated with the acquired credential handle. 

We already pass certificates to `AcquireCredentialHandle` so this behavior should
also work out of the box (not sure if client auth is/was used before).

sfackler:
Ah, right, misread that check. LGTM

PR #26, opened 2017-05-23
Add a verification callback.

steffengy:
Openssl-inspired implementation of verification callbacks.

TODOs as of now:
- [x] Lifetime's: CertChain should likely borrow/clone CertChainContext!
When the related CertChainContext goes out of scope, CertChain points to invalid memory!
- [x] Tests
- [x] ~~Better solution for the `Arc<Fn...>` ?~~

### DIFF
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -9,10 +9,11 @@ use std::io::{self, Read, BufRead, Write, Cursor};
 use std::mem;
 use std::ptr;
 use std::slice;
+use std::sync::Arc;
 use winapi;
 
 use {INIT_REQUESTS, ACCEPT_REQUESTS, Inner, secbuf, secbuf_desc};
-use cert_context::CertContext;
+use cert_chain::{CertChain, CertChainContext};
 use cert_store::CertStore;
 use security_context::SecurityContext;
 use context_buffer::ContextBuffer;
@@ -27,20 +28,11 @@ lazy_static! {
         winapi::szOID_SGC_NETSCAPE.bytes().chain(Some(0)).collect();
 }
 
-struct CertChainContext(winapi::PCERT_CHAIN_CONTEXT);
-
-impl Drop for CertChainContext {
-    fn drop(&mut self) {
-        unsafe {
-            crypt32::CertFreeCertificateChain(self.0);
-        }
-    }
-}
-
 /// A builder type for `TlsStream`s.
-#[derive(Default, Debug)]
+#[derive(Default)]
 pub struct Builder {
     domain: Option<Vec<u16>>,
+    verify_callback: Option<Arc<Fn(bool, &CertChain) -> bool>>,
     cert_store: Option<CertStore>,
     accept: bool,
 }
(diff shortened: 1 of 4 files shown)


steffengy:
@sfackler
Any ideas/comments on the todos/implementation?

Any alternative to using `Arc<Fn(bool, &CertChain) -> bool>`  ?

src/tls_stream.rs
```diff
@@ -60,6 +52,17 @@ impl Builder {
         self
     }
 
+    /// Set a verification callback to be used for connections created with this `Builder`.
+    ///
+    /// The callback is provided with a boolean indicating if the (pre)validation was
+    /// successful and with the certificate, which was validated.
+    pub fn verify_callback<F>(&mut self, callback: F) -> &mut Builder 
+        where F: Fn(bool, &CertChain) -> bool + 'static
```
sfackler: I think we may want to swap the `bool`s out for `io::Result<()>` - it allows the closure to inspect the upstream validation error and pick which error they want to produce.

steffengy: Sounds reasonable. Done.

steffengy:
> I think Arc<Fn>'s just fine honestly. It could be a special VerifyChain trait that has a blanket impl for the appropriate closure type I guess?

Okay. Is there any substantial advantage for the `VerifyChain` solution? I personally also
think `Arc<Fn>` is fine for 99% of the usage scope.

sfackler:
Nothing particularly concrete I don't think. It does allow you to add new (defaulted) methods in the future, but I have no idea why we'd want to do that here.

steffengy:
Okay so I'll guess I'll leave it at that.
Anything else before merging?

src/tls_stream.rs
```diff
@@ -60,6 +52,17 @@ impl Builder {
         self
     }
 
+    /// Set a verification callback to be used for connections created with this `Builder`.
+    ///
+    /// The callback is provided with a boolean indicating if the (pre)validation was
```
sfackler: Docs need to be updated.

sfackler:
Looks good other than the docs update!

steffengy:
Okay docs are updated and will be merged soon.

alexcrichton:
@steffengy I think this may have [indirectly broken](https://ci.appveyor.com/project/alexcrichton/tokio-tls/build/1.0.108) the tokio-tls CI. I think it's due to the trait object no longer being `Send`? Would it be possible to add Send/Sync bounds here to restore the send/sync-ness of the types?

(or is this totally the wrong change?)

steffengy:
@alexcrichton 
`CertValidationResult` should implement Send/Sync for sure.
I'll look into it and check if that resolves the CI-error for tokio-tls.
Thanks for the report!

src/tls_stream.rs
```diff
@@ -27,20 +28,11 @@ lazy_static! {
...
-        }
-    }
-}
-
 /// A builder type for `TlsStream`s.
-#[derive(Default, Debug)]
+#[derive(Default)]
 pub struct Builder {
     domain: Option<Vec<u16>>,
+    verify_callback: Option<Arc<Fn(io::Result<()>, &CertChain) -> io::Result<()>>>,
```
sfackler: And here

src/tls_stream.rs
```diff
@@ -60,6 +52,19 @@ impl Builder {
...
     }
 
+    /// Set a verification callback to be used for connections created with this `Builder`.
+    ///
+    /// The callback is provided with an io::Result indicating if the (pre)validation was  
+    /// successful. The Ok() variant indicates a successful validation while the Err() variant  
+    /// contains the errorcode returned from the internal verification process.    
+    /// The validated certificate, is accessible through the second argument of the closure.
+    pub fn verify_callback<F>(&mut self, callback: F) -> &mut Builder 
+        where F: Fn(io::Result<()>, &CertChain) -> io::Result<()> + 'static
```
sfackler: Oh yeah, we need to add `+ Sync + Send` here

steffengy:
@alexcrichton 
Fixed in 0.1.7 https://github.com/steffengy/schannel-rs/commit/7f1f1be0fb7dbea4a75f612db45977be78986a31

alexcrichton:
Thanks @steffengy!

PR #27, opened 2017-05-25
Add more generic certificate fingerprinting API

steffengy:
This adds a similar interface to openssl's in regards to certificate fingerprinting.
Only SHA224 and RIPEMD160 are not supported since they aren't supported
by crypto32.

### DIFF
--- a/src/cert_context.rs
+++ b/src/cert_context.rs
@@ -28,13 +28,29 @@ unsafe impl Sync for CertContext {}
 unsafe impl Send for CertContext {}
 
 /// A supported hashing algorithm
+pub struct HashAlgorithm(winapi::DWORD, usize);
+
 #[allow(missing_docs)]
-pub enum HashAlgorithm {
-    MD5,
-    SHA1,
-    SHA256,
-    SHA384,
-    SHA512,
+impl HashAlgorithm {
+    pub fn md5() -> HashAlgorithm {
+        HashAlgorithm(winapi::CALG_MD5, 16)
+    }
+
+    pub fn sha1() -> HashAlgorithm{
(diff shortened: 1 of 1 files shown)


src/cert_context.rs
```diff
@@ -76,19 +86,27 @@ impl CertContext {
         }
     }
 
-    /// Returns the sha1 hash of this certificate
-    ///
-    /// The sha1 is returned as a 20-byte array representing the bits of the
-    /// sha1 hash.
-    pub fn sha1(&self) -> io::Result<[u8; 20]> {
```
sfackler: We can't remove this outright without breaking back compat.

src/cert_context.rs
```diff
@@ -27,6 +27,16 @@ pub struct CertContext(winapi::PCCERT_CONTEXT);
 unsafe impl Sync for CertContext {}
 unsafe impl Send for CertContext {}
 
+/// A supported hashing algorithm
+#[allow(missing_docs)]
+pub enum HashAlgorithm {
```
sfackler: I don't really like using enums for these kinds of things - it's not extensible to new algorithms as-is and no-one ever really wants to match on it. I like using an opaque wrapper around the underlying `DWORD` or whatever it is.

src/cert_context.rs
steffengy: Right. I actually did want to add it as wrapper for fingerprint, seems like I didn't.
Anyways fixed now.

src/cert_context.rs
steffengy: If I understood you correctly, it should be fixed now.

steffengy:
Anything else here that should be addressed before a merge?

PR #28, opened 2017-06-01
pass security context up on handshake error

sstecko:
Adds new HandshakeError::Failure2 enum entry that also provides the failed securitycontext. This will allow people to present failed certificates for validation, etc.

### DIFF
--- a/src/security_context.rs
+++ b/src/security_context.rs
@@ -10,6 +10,7 @@ use context_buffer::ContextBuffer;
 
 use schannel_cred::SchannelCred;
 
+#[derive(Debug)]
 pub struct SecurityContext(winapi::CtxtHandle);
 
 impl Drop for SecurityContext {
@@ -38,7 +39,7 @@ impl SecurityContext {
     pub fn initialize(cred: &mut SchannelCred,
                       accept: bool,
                       domain: Option<&[u16]>)
-                      -> io::Result<(SecurityContext, Option<ContextBuffer>)> {
+                      -> Result<(SecurityContext, Option<ContextBuffer>),(io::Error, SecurityContext)> {
         unsafe {
             let mut ctxt = mem::zeroed();
 
@@ -71,7 +72,7 @@ impl SecurityContext {
                     Ok((SecurityContext(ctxt), Some(ContextBuffer(outbuf[0]))))
                 }
                 err => {
-                    Err(io::Error::from_raw_os_error(err as i32))
+                    Err((io::Error::from_raw_os_error(err as i32), SecurityContext(ctxt)))
                 }
             }
         }
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -79,14 +79,8 @@ impl Builder {
         self
     }
 
-    /// Initialize a new TLS session where the stream provided will be
-    /// connecting to a remote TLS server.
-    ///
-    /// If the stream provided is a blocking stream then the entire handshake
-    /// will be performed if possible, but if the stream is in nonblocking mode
-    /// then a `HandshakeError::Interrupted` variant may be returned. This
-    /// type can then be extracted to later call
-    /// `MidHandshakeTlsStream::handshake` when data becomes available.
+    ///DEPRECATED this connect will not return a HandshakeError::Failure2
+    #[deprecated(note = "please use connect2 instead")]
     pub fn connect<S>(&mut self,
                       cred: SchannelCred,
                       stream: S)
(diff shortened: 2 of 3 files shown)


src/security_context.rs
```diff
@@ -71,7 +72,7 @@ impl SecurityContext {
                     Ok((SecurityContext(ctxt), Some(ContextBuffer(outbuf[0]))))
                 }
                 err => {
-                    Err(io::Error::from_raw_os_error(err as i32))
+                    Err((io::Error::from_raw_os_error(err as i32), SecurityContext(ctxt)))
```
sfackler: The context object isn't valid if `InitializeSecurityContextW` fails, so we shouldn't be returning it here.

src/tls_stream.rs
```diff
@@ -116,6 +127,8 @@ impl Builder {
         self.initialize(cred, true, stream)
     }
 
+    /// DEPRECATED this initialize function will not return a HandshakeError::Failure2
+    #[deprecated(note = "please use initialize2 instead")]
```
sfackler: This is a private method, so it doesn't need to be deprecated.

src/tls_stream.rs
```diff
@@ -79,6 +79,17 @@ impl Builder {
...
 
+    ///DEPRECATED this connect will not return a HandshakeError::Failure2
+    #[deprecated(note = "please use connect2 instead")]
+    pub fn connect<S>(&mut self,
+                      cred: SchannelCred,
+                      stream: S)
+                      -> Result<TlsStream<S>, HandshakeError<S>>
+        where S: Read + Write
+    {
+        self.initialize(cred, false, stream)
```
sfackler: Can we just have this delegate to connect2 and adjust the error type to avoid having to duplicate a bunch of stuff?

src/tls_stream.rs
```diff
@@ -195,6 +249,10 @@ pub struct TlsStream<S> {
 pub enum HandshakeError<S> {
     /// A fatal I/O error occurred
     Failure(io::Error),
+    
+    // Fatal I/O error, with security context.
+    Failure2(io::Error, SecurityContext),
```
sfackler: Adding this variant breaks backwards compatibility - we'll need to make a new enum.

src/security_context.rs
sstecko: That's fair, but this is the only sensical place I could see to grab it because it seems that it needs to be initialized one way or another to hold meaningful failure information. The Handshake failure happens right here as far as I can tell. Maybe the SecurityContext isn't the actual object we want to be passing back in case of a failure?

src/tls_stream.rs
sstecko: sure good idea

src/tls_stream.rs
sstecko: k

src/tls_stream.rs
sstecko: bummer but makes sense

steffengy:
Just to make sure there's an actual use case to this (Which I guess there is?):

What information can you get from the security context for a failed verification
that you can't get from the certificate chain using the verify_callback?

sstecko:
Does the verify_callback certChain give any clue as to which certificate failed to validate in the case of a failed handshake? The main use case I'm trying to enable is:

1. Application attempts to make secure connection, fails
2. User is prompted with error that explains why and prompts for override (eg. "Error: Self signed certificate found in certificate chain. Fingerprint: 1039e91F10a183be1240ff would you like to trust this certificate?")
3. User trusts certificate, it is added to user's truststore and the connection is retried. 

The only issue I see with the current Verify_callback method is that the error is not very descriptive (io, basically "failed to open stream" if I'm not mistaken), and the cert chain gives no clue as to which certificate actually failed to validate. 

Of course I could be mistaken

steffengy:
@sstecko 
No it doesn't give any hint about what went wrong during validation.

That information AFAIK is only available in [`_CERT_CHAIN_POLICY_STATUS`](https://github.com/steffengy/schannel-rs/blob/cd3929a47ad993b3435ab584223f024ee7ac88de/src/tls_stream.rs#L556),
so I don't see how exposing the `SecurityContext` is helping in case of this problem?
If I'm missing something there, please clarify.

sstecko:
I must have misunderstood the purpose of the SecurityContext. I will open another PR with changes to verify_callback that pass up additional information to the closure besides the cert chain. 

Before I start, do you think it would be appropriate to pass up the status object from the validation function, or should I wrap it somehow?

steffengy:
```c
typedef struct _CERT_CHAIN_POLICY_STATUS {
 DWORD cbSize;                // unnecessary
 DWORD dwError;               // passed as Err() of the io::Result
 LONG  lChainIndex;           // relevant
 LONG  lElementIndex;         // relevant
 void  *pvExtraPolicyStatus;  // not relevant for CERT_CHAIN_POLICY_SSL (but maybe long-term)
} CERT_CHAIN_POLICY_STATUS, *PCERT_CHAIN_POLICY_STATUS;
```

My first idea would be adding something like a `CertValidationResult`
which wraps **both** the chain and the policy status above.
(We also need a `chain` function implemented for it to access the wrapped chain)

The advantage would be that you could implement `get_failed_certificate()`
which would give you the certificate specified by `lChainIndex` and `lElement`,
when applicable.

The error code/reason should be accessible through the io::Result
(which is the same error code as in `.dwError` and could e.g. be [`CERT_E_EXPIRED`](https://msdn.microsoft.com/ru-ru/aa377188)

NOTE: The names might need some improvement

sstecko:
Closed in favor of https://github.com/steffengy/schannel-rs/pull/29/files

PR #29, opened 2017-06-06
variant of verify_callback with additional context

sstecko:
This enables the workfow where an individual needs user confirmation for a failed certificate during verification callback.

### DIFF
--- a/src/cert_chain.rs
+++ b/src/cert_chain.rs
@@ -2,6 +2,7 @@
 
 use std::mem;
 use std::slice;
+use std::vec::Vec;
 use crypt32;
 use winapi;
 
@@ -34,14 +35,27 @@ impl CertChainContext {
     /// https://msdn.microsoft.com/de-de/library/windows/desktop/aa377182(v=vs.85).aspx
     /// rgpChain[cChain - 1] is the final chain
     pub fn final_chain(&self) -> Option<CertChain> {
-        let cloned = self.clone();
-        let chains = unsafe {
-            let cert_chain = *cloned.0;
+        if let Some(chain) = self.chains().last(){
+            return Some(CertChain(chain.0, self.clone()));
+        }
+        None
+    }
+    
+    pub fn get_chain(&self, index :usize) -> Option<CertChain> {
+        if let Some(chain) = self.chains().get(index) {
+            return Some(CertChain(chain.0, self.clone()));
+        }
+        None
+    }
+    
+    pub fn chains(&self) -> Vec<CertChain> {
+        let cert_chains = unsafe {
+            let cert_chain = *self.0;
             slice::from_raw_parts(
                 cert_chain.rgpChain as *mut winapi::PCERT_SIMPLE_CHAIN,
                 cert_chain.cChain as usize)
         };
-        chains.last().map(|chain| CertChain(*chain, cloned))
+        return cert_chains.iter().map(|chain| CertChain(*chain, self.clone())).collect();
     }
 }
 
--- a/src/cert_chain.rs
+++ b/src/cert_chain.rs
@@ -2,7 +2,6 @@
 
 use std::mem;
 use std::slice;
-use std::vec::Vec;
 use crypt32;
 use winapi;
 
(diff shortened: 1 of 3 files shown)


src/tls_stream.rs
```diff
@@ -58,13 +60,27 @@ impl Builder {
...
         self
     }
 
+    /// Set a verification callback to be used for connections created with this `Builder`.
+    ///
+    /// The callback is provided with an io::Result indicating if the (pre)validation was  
+    /// successful. The Ok() variant indicates a successful validation while the Err() variant  
+    /// contains the errorcode returned from the internal verification process.    
+    /// The validated certificate, is accessible through the second argument of the closure.
+    pub fn verify_callback2<F>(&mut self, callback: F) -> &mut Builder
```
sfackler: I don't think we've cut a release with `verify_callback` so we can just update this in place.

src/cert_chain.rs
```diff
@@ -34,14 +35,27 @@ impl CertChainContext {
...
+    }
+    
+    pub fn get_chain(&self, index :usize) -> Option<CertChain> {
+        if let Some(chain) = self.chains().get(index) {
+            return Some(CertChain(chain.0, self.clone()));
+        }
+        None
+    }
+    
+    pub fn chains(&self) -> Vec<CertChain> {
```
sfackler: This should probably return an iterator rather than a vec.

src/cert_chain.rs
```diff
@@ -34,14 +35,27 @@ impl CertChainContext {
...
-        let cloned = self.clone();
-        let chains = unsafe {
-            let cert_chain = *cloned.0;
+        if let Some(chain) = self.chains().last(){
+            return Some(CertChain(chain.0, self.clone()));
+        }
+        None
+    }
+    
+    pub fn get_chain(&self, index :usize) -> Option<CertChain> {
```
sfackler: We'll want some way to know how many chains there are if we have this, and the implementation shouldn't be a linear scan.

src/tls_stream.rs
```diff
@@ -200,6 +218,35 @@ pub enum HandshakeError<S> {
...
+    res :io::Result<()>,
+    chain_index :i32,
+    element_index :i32,
+    extra_policy_status :Option<winapi::CERT_CHAIN_POLICY_STATUS>,
+}
+
+impl CertValidationResult {
+
+    /// Returns the certificate that failed validation if applicable
+    pub fn get_failed_certificate(&self) -> Option<CertContext> {
```
sfackler: Getter functions typically omit the `get_` prefix.

sstecko:
Attempted to make the changes SFackler suggested. Thanks for the review

shadytradesman:
@steffengy @sfackler I'm not sure why this check is failing, but I don't think it's due to my code change.

sstecko:
@steffengy[github.com] @sfackler[github.com] I'm not sure why this check is failing, but I don't think it's due to my code change.

steffengy:
It's just a nightly regression so that currently GCC will not be found.
https://github.com/rust-lang/rust/issues/42422

So you can safely ignore it, in the next few days it'll fix itself.

sstecko:
Hey guys, is there anything I need to do to push this PR forward? I'd really like to get this merge and have a release cut so I can use this with hyper-native-tls in a project :)

src/tls_stream.rs
```diff
@@ -32,7 +33,7 @@ lazy_static! {
 #[derive(Default)]
 pub struct Builder {
     domain: Option<Vec<u16>>,
-    verify_callback: Option<Arc<Fn(io::Result<()>, &CertChain) -> io::Result<()>>>,
+    verify_callback: Option<Arc<Fn(CertValidationResult) -> io::Result<()>>>,
```
sfackler: This needs to be passed a reference, not a result by value or it'll have access to a bunch of dangling pointers once the callback returns.

sfackler: Oh maybe not...

src/tls_stream.rs
```diff
@@ -200,6 +201,35 @@ pub enum HandshakeError<S> {
     Interrupted(MidHandshakeTlsStream<S>),
 }
 
+/// A struct used to wrap various cert chain validation results for callback processing. 
+pub struct CertValidationResult {
+    chain :CertChainContext,
+    res :io::Result<()>,
+    chain_index :i32,
+    element_index :i32,
+    extra_policy_status :Option<winapi::CERT_CHAIN_POLICY_STATUS>,
```
sfackler: Looks like this is unused?

src/cert_chain.rs
sfackler: Can we cut the `get_` prefix off of this?

src/tls_stream.rs
```diff
@@ -200,6 +201,35 @@ pub enum HandshakeError<S> {
...
+        }
+        None
+    }
+    
+    // Returns the final certificate chain in the certificate context if applicable
+    pub fn chain(&self) -> Option<CertChain> {
+        self.chain.final_chain()
+    }
+    
+    pub fn result(&self) -> &io::Result<()> {
```
sfackler: `io::Error` isn't `Clone`, so we'll want this to return the `Result` by value so that it can then be returned from the closure - we should probably store the error code directly and construct the `Error` on each call.

src/tls_stream.rs
```diff
@@ -200,6 +201,35 @@ pub enum HandshakeError<S> {
     Interrupted(MidHandshakeTlsStream<S>),
 }
 
+/// A struct used to wrap various cert chain validation results for callback processing. 
+pub struct CertValidationResult {
+    chain :CertChainContext,
+    res :io::Result<()>,
+    chain_index :i32,
+    element_index :i32,
+    extra_policy_status : (winapi::LPCSTR, *const winapi::c_void),
```
sfackler: Can we get rid of these fields if they aren't used?

sfackler: I don't think it'd be breaking - they're not publicly accessible, right?

sstecko:
Hey @sfackler I have put up another revision. I appreciate your attention here, as we are somewhat blocked on getting these changes merged in and incorporated into the project.

src/tls_stream.rs
steffengy: To clarify: 
I only meant to describe which struct fields could potentially be relevant for us.
Those particular fields might be relevant for some use case in the future 
(when we would access them and expose them without breaking BC).

Since we currently aren't even aware of a use case,
there's no need to keep them around.

sfackler:
LGTM, thanks! I won't be able to publish until I get home to my Windows box tonight, but you should be able to unblock yourself by doing a package replacement: http://doc.crates.io/manifest.html#the-replace-section.

steffengy:
@sfackler 
I'll look into getting this merged & published within a few minutes.

steffengy:
Published as 0.1.6.

PR #30, opened 2017-06-21
restore BC: the validation callback (+TlsStream) should be Sync/Send,…

steffengy:
Just fixing some missed details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "schannel"
 license = "MIT"
-version = "0.1.6"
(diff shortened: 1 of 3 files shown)


sfackler:
Looks good to me. Might be worth adding in some little checks to keep this from regressing in the future:

```
fn _is_sync() {
    fn sync<T: Sync>() {}
    sync::<TlsStream<()>>();
}
```

sstecko:
hah. Just noticed this when trying to update rust-native-tls. Thanks

PR #31, opened 2017-06-23
Support loading identities from key/cert files

sfackler:
This unblocks https://github.com/sfackler/rust-native-tls/issues/27!


src/test.rs
```diff
@@ -558,3 +559,68 @@ fn accept_one_byte_at_a_time() {
...
+        stream.flush().unwrap();
+        assert_eq!(stream.read(&mut [0; 1024]).unwrap(), 4);
+        stream.shutdown().unwrap();
+    });
+
+    let cert = include_bytes!("../test/cert.der");
+    let cert = CertContext::new(cert).unwrap();
+
+    let mut options = AcquireOptions::new();
+    options.container("schannel-test");
```
jethrogb: If I understand CryptAcquireContext correctly, if you don't use `CRYPT_VERIFYCONTEXT`, the key is stored permanently on disk, is that right?

steffengy: [MSDN/CryptAcquireContext](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379886(v=vs.85).aspx):
>> Instead, either prevent key storage by passing the CRYPT_VERIFYCONTEXT flag in the dwFlags parameter, or use an application-specific container that is unlikely to be used by another application. [...] For performance reasons, we recommend that you set the pszContainer parameter to NULL and the dwFlags parameter to CRYPT_VERIFYCONTEXT in all situations where you do not require a persisted key.

That sounds to me like we want `verify_context()` here.

sfackler:
You can't decrypt data with a key in a verify context. Those are used for
things like signature verification but won't work for TLS.

Issue #34, opened 2017-07-14
Something is wrong with TlsStream's Write implementation leading to data corruption


e00E:
I noticed data corruption when sending data over https with reqwest and started investigating: https://github.com/seanmonstar/reqwest/issues/164

First I saw that sometimes `TlsStream`'s `encrypt` method would never receive the last byte of my input.
Then I looked at `TlsStream`'s `Write` implementation at https://github.com/steffengy/schannel-rs/blob/master/src/tls_stream.rs#L798 and noticed the following behavior:
* `write` is called with a buffer of length 106, this buffer has at its end the second to last byte of my input
* `write` enters the if and calls encrypt on this buffer
* `write` calls `write_out` and forwards the resulting `WouldBlock`
* `write` is called twice with a buffer of length107, now also containing the very last byte of my input at its end
* both times `write` does not call `encrypt` but calls `write_out` and forwards the resulting `WouldBlock`
* `write` is called a third time again with that buffer, it again does not call `encrypt` but calls` write_out `successfully (which returned` Ok(135)`) and returns `Ok(107)`

I dont know what exactly is going wrong yet or how to fix it but this definitely looks like a bug in`schannel-rs` to me. `TlsStream` is receiving a buffer to write and responds that it has written the buffer while not having ever encrypted or sent the last byte of that buffer.

e00E:
I think the comment in `write` describes a erroneous assumption:
> if we have pending output data, it must have been because a previous attempt to send this data ran into an error. Specifically in the case of WouldBlock errors, **we expect another call to write with the same data**.

This is not guaranteed and caller has no way of knowing that, they just sees a standard Write implementation.
This is also what goes wrong in my example above. There is a WouldBlock error and in the next call the input buffer has changed, it has only changed but by one additional byte in this case but really it could be anything.
Changing input buffers between calls seem to cause a disconnect of the contents of `out_buf` and the actual input the to `write`.

e00E:
I think this sentence from the Write trait is also violated `A call to write represents at most one attempt to write to any wrapped object.` by repeatedly writing to `self.stream` in a loop in `write_out`.

e00E:
Also couldn't `write_out` enter an infinite loop if `self.stream.write` started returning 0.

sfackler:
Why is more data being added to the buffer being passed to `write` in between calls?

e00E:
I dont know. Reqwests uses hyper so probably there?
[...]

PR #35, opened 2017-07-14
Fix wrong Write implementation.

e00E:
Fixes https://github.com/steffengy/schannel-rs/issues/34

I believe this is a solution which is both efficient by not discarding earlier encrypted bytes and correct by abiding the rules of the `Write` trait.

### DIFF
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -617,11 +617,15 @@ impl<S> TlsStream<S>
     }
 
     fn write_out(&mut self) -> io::Result<usize> {
-        let mut nwritten = 0;
-		let position = self.out_buf.position() as usize;
-		nwritten += self.stream.write(&self.out_buf.get_ref()[position..])?;
-		self.out_buf.set_position((position + nwritten) as u64);
-        Ok(nwritten)
+        let mut out = 0;
+        while self.out_buf.position() as usize != self.out_buf.get_ref().len() {
+            let position = self.out_buf.position() as usize;
+            let nwritten = try!(self.stream.write(&self.out_buf.get_ref()[position..]));
+            out += nwritten;
+            self.out_buf.set_position((position + nwritten) as u64);
+        }
+
+        Ok(out)
 	}
 
     fn read_in(&mut self) -> io::Result<usize> {
@@ -798,26 +802,27 @@ impl<S> Write for TlsStream<S>
 		
 		// We can only write if the write buffer is emptied first
 		self.write_out()?;
-        // Check if write_out managed to write the whole buffer or if some bytes are remaining
-		if self.out_buf.position() as usize != self.out_buf.get_ref().len() {
-			return Err(io::Error::new(io::ErrorKind::WouldBlock, "write buffer was not empty"))
-		}
 
         let len = cmp::min(buf.len(), sizes.cbMaximumMessage as usize);
 
-        self.out_buf = Cursor::new(self.encrypt(&buf[..len], &sizes)?);
+        let mut encrypted = self.encrypt(&buf[..len], &sizes)?;
+        self.out_buf.set_position(0);
+        self.out_buf.get_mut().clear(); // Does not change allocated capacity
+        self.out_buf.get_mut().append(&mut encrypted);
+        // TODO: not sure if this should be there
+        // it would use less memory but reallocate more often
+        // self.out_buf.get_mut().shrink_to_fit();
 
-		// Pretend we wrote everything because we put it on the write buffer
+        // TODO: we could call self.write_out again here to see if we can immediately
+        // write out more. Should we?
+
+        // Pretend we wrote everything because we put it on the write buffer
         Ok(len)
     }
 
     fn flush(&mut self) -> io::Result<()> {
 		// Make sure the write buffer is emptied
 		self.write_out()?;
-        // Check if write_out managed to write the whole buffer or if some bytes are remaining
-		if self.out_buf.position() as usize != self.out_buf.get_ref().len() {
-			return Err(io::Error::new(io::ErrorKind::WouldBlock, "write was buffer not empty"))
-		}
         self.stream.flush()
     }
 }
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -821,8 +821,8 @@ impl<S> Write for TlsStream<S>
     }
 
     fn flush(&mut self) -> io::Result<()> {
-		// Make sure the write buffer is emptied
-		self.write_out()?;
+        // Make sure the write buffer is emptied
+        self.write_out()?;
         self.stream.flush()
     }
 }
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -619,9 +619,8 @@ impl<S> TlsStream<S>
     fn write_out(&mut self) -> io::Result<usize> {
         let mut nwritten = 0;
 		let position = self.out_buf.position() as usize;
-		nwritten += try!(self.stream.write(&self.out_buf.get_ref()[position..]));
+		nwritten += self.stream.write(&self.out_buf.get_ref()[position..])?;
 		self.out_buf.set_position((position + nwritten) as u64);
-
         Ok(nwritten)
 	}
 
@@ -799,8 +798,9 @@ impl<S> Write for TlsStream<S>
 		
 		// We can only write if the write buffer is emptied first
 		self.write_out()?;
+        // Check if write_out managed to write the whole buffer or if some bytes are remaining
 		if self.out_buf.position() as usize != self.out_buf.get_ref().len() {
-			return Err(io::Error::new(io::ErrorKind::WouldBlock, "write buffer not empty"))
+			return Err(io::Error::new(io::ErrorKind::WouldBlock, "write buffer was not empty"))
 		}
 
         let len = cmp::min(buf.len(), sizes.cbMaximumMessage as usize);
@@ -812,10 +812,11 @@ impl<S> Write for TlsStream<S>
     }
 
     fn flush(&mut self) -> io::Result<()> {
-		// Make sure pending writes are done
+		// Make sure the write buffer is emptied
 		self.write_out()?;
+        // Check if write_out managed to write the whole buffer or if some bytes are remaining
 		if self.out_buf.position() as usize != self.out_buf.get_ref().len() {
-			return Err(io::Error::new(io::ErrorKind::WouldBlock, "write buffer not empty"))
+			return Err(io::Error::new(io::ErrorKind::WouldBlock, "write was buffer not empty"))
 		}
         self.stream.flush()
     }
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -733,11 +733,7 @@ impl<S> TlsStream<S>
         let len = sizes.cbHeader as usize + buf.len() + sizes.cbTrailer as usize;
 
         self.out_buf.set_position(0);
-        if self.out_buf.get_ref().len() < len {
-            self.out_buf.get_mut().resize(len, 0);
-        } else {
-            self.out_buf.get_mut().truncate(len); // Does not change capacity
-        }
+        self.out_buf.get_mut().resize(len, 0);
 
         let message_start = sizes.cbHeader as usize;
         self.out_buf
@@ -762,6 +758,8 @@ impl<S> TlsStream<S>
 
             match secur32::EncryptMessage(self.context.get_mut(), 0, &mut bufdesc, 0) {
                 winapi::SEC_E_OK => {
+                    let len = bufs[0].cbBuffer + bufs[1].cbBuffer + bufs[2].cbBuffer;
+                    self.out_buf.get_mut().truncate(len as usize);
                     Ok(())
                 }
                 err => Err(io::Error::from_raw_os_error(err as i32)),
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -617,16 +617,13 @@ impl<S> TlsStream<S>
     }
 
     fn write_out(&mut self) -> io::Result<usize> {
-        let mut out = 0;
-        while self.out_buf.position() as usize != self.out_buf.get_ref().len() {
-            let position = self.out_buf.position() as usize;
-            let nwritten = try!(self.stream.write(&self.out_buf.get_ref()[position..]));
-            out += nwritten;
-            self.out_buf.set_position((position + nwritten) as u64);
-        }
+        let mut nwritten = 0;
+		let position = self.out_buf.position() as usize;
+		nwritten += try!(self.stream.write(&self.out_buf.get_ref()[position..]));
+		self.out_buf.set_position((position + nwritten) as u64);
 
-        Ok(out)
-    }
+        Ok(nwritten)
+	}
 
     fn read_in(&mut self) -> io::Result<usize> {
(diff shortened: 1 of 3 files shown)


e00E:
Firstly this PR changes `write_out` to comply with `A call to write represents at most one attempt to write to any wrapped object.` 

Secondly it changes `write` to always first try to `write_out` the write buffer. If the buffer could not be emptied `WouldBlock` is returned.
If the write buffer could be emptied, it accepts the next bytes to send by encrypting them and storing them in the write buffer and then pretending to have written all bytes already.
This works because the next call to write will write the previously buffered bytes out before accepting new ones. The very last write will be written when `flush` is called.

It also changes `encrypt` to return the encrypted bytes instead of appending them to the write buffer right there because that makes it more clear how `write` works and what assumptions it makes about the write buffer.

src/tls_stream.rs
```diff
@@ -617,16 +617,12 @@ impl<S> TlsStream<S>
...
-            let nwritten = try!(self.stream.write(&self.out_buf.get_ref()[position..]));
-            out += nwritten;
-            self.out_buf.set_position((position + nwritten) as u64);
-        }
-
-        Ok(out)
-    }
+        let mut nwritten = 0;
+		let position = self.out_buf.position() as usize;
+		nwritten += self.stream.write(&self.out_buf.get_ref()[position..])?;
```
sfackler: The code below expects that the entire out_buf is written out by this method, but it doesn't loop so that won't necessarily happen.

src/tls_stream.rs
```diff
@@ -803,22 +795,29 @@ impl<S> Write for TlsStream<S>
...
 
-        // if we have pending output data, it must have been because a previous
-        // attempt to send this data ran into an error. Specifically in the
-        // case of WouldBlock errors, we expect another call to write with the
-        // same data.
-        if self.out_buf.position() == self.out_buf.get_ref().len() as u64 {
-            try!(self.encrypt(&buf[..len], &sizes));
-        }
-        try!(self.write_out());
+        self.out_buf = Cursor::new(self.encrypt(&buf[..len], &sizes)?);
```
sfackler: It seems bad to be allocating a new buffer on every write.

src/tls_stream.rs
```diff
@@ -617,16 +617,12 @@ impl<S> TlsStream<S>
...
-            let position = self.out_buf.position() as usize;
-            let nwritten = try!(self.stream.write(&self.out_buf.get_ref()[position..]));
-            out += nwritten;
-            self.out_buf.set_position((position + nwritten) as u64);
-        }
-
-        Ok(out)
-    }
+        let mut nwritten = 0;
+		let position = self.out_buf.position() as usize;
```
sfackler: There appear to be hard tabs here.

src/tls_stream.rs
e00E: Ah if other code expects write_out to do that then I guess we have to change it back. Like I said above this was specifically changed to abide by `A call to write represents at most one attempt to write to any wrapped object.` . Maybe we can change the other code that depends on that behavoir? I dont know the surrounding code that well.

src/tls_stream.rs
e00E: Whops. I was wondering why it looked strange in the diff, not sure how I managed to insert that.

src/tls_stream.rs
steffengy: > A call to write represents at most one attempt to write to any wrapped object.

This isn't well formulated in the current docs, but the intention of it is
of `Write` to be as low level as the implementation allows.
Which is exactly what it is in that case (so it's fine to do repeated writes there).

https://internals.rust-lang.org/t/std-io-write-a-call-to-write-represents-at-most-one-attempt-to-write-to-any-wrapped-object/5068

e00E: Thanks, I will change it back.

sfackler: The general rule of thumb is to respect that rule as much as you can. We can't do that here since forcing a single write in this manner means the stream totally breaks if the kernel buffer happens to fill.

e00E:
I fixed some tests too, which also failed on master I think.
`test test::valid_algorithms has been running for over 60 seconds` I did not fix because I dont know what to do.

e00E:
Also I marked two things as `TODO` in `write` where I wasn't what the best approach was, if you don't mind looking at that again.

steffengy:
All tests pass on master:
https://ci.appveyor.com/project/steffengy/schannel-rs/branch/master

> test test::valid_algorithms has been running for over 60 seconds

That sounds like a bug caused by this. 
Likely some infinite loop somewhere
(or some test code calling into schannel and waiting for it to get something done
and it never gets done)

e00E:
Thats interresting. The tests failed for me not beacuse of this PR but because google returned a moved status code instead of success.

e00E:
Your link is 20 days old, I think they dont work anymore.

e00E:
I tried to fix `test::valid_algorithms` in the same way I fixed the others but it still fails:
```
---- test::valid_algorithms stdout ----
        thread 'test::valid_algorithms' panicked at 'called `Result::unwrap()` on an `Err` value: Failure(Error { re
pr: Os { code: -2146893018, message: "The message received was unexpected or badly formatted." } })', src\libcore\re
sult.rs:859
stack backtrace:
   0: std::sys_common::backtrace::_print
             at C:\projects\rust\src\libstd\sys_common\backtrace.rs:94
   1: std::panicking::default_hook::{{closure}}
             at C:\projects\rust\src\libstd\panicking.rs:354
   2: std::panicking::default_hook
             at C:\projects\rust\src\libstd\panicking.rs:365
   3: std::panicking::rust_panic_with_hook
             at C:\projects\rust\src\libstd\panicking.rs:549
   4: std::panicking::begin_panic<collections::string::String>
             at C:\projects\rust\src\libstd\panicking.rs:511
   5: std::panicking::begin_panic_fmt
             at C:\projects\rust\src\libstd\panicking.rs:495
   6: std::panicking::rust_begin_panic
             at C:\projects\rust\src\libstd\panicking.rs:471
   7: core::panicking::panic_fmt
             at C:\projects\rust\src\libcore\panicking.rs:69
   8: core::result::unwrap_failed<schannel::tls_stream::HandshakeError<std::net::tcp::TcpStream>>
             at C:\projects\rust\src\libcore\macros.rs:29
   9: core::result::Result<schannel::tls_stream::TlsStream<std::net::tcp::TcpStream>, schannel::tls_stream::Handshak
eError<std::net::tcp::TcpStream>>::unwrap<schannel::tls_stream::TlsStream<std::net::tcp::TcpStream>,schannel::tls_st
ream::HandshakeError<std::net::tcp::TcpStream>>
             at C:\projects\rust\src\libcore\result.rs:737
  10: schannel::test::valid_algorithms
             at .\src\test.rs:52
  11: test::{{impl}}::call_box<(),closure>
             at C:\projects\rust\src\libtest\lib.rs:140
  12: panic_unwind::__rust_maybe_catch_panic
             at C:\projects\rust\src\libpanic_unwind\lib.rs:98
  13: std::panicking::try::do_call<std::panic::AssertUnwindSafe<closure>,()>
             at C:\projects\rust\src\libstd\panicking.rs:450
  14: panic_unwind::__rust_maybe_catch_panic
             at C:\projects\rust\src\libpanic_unwind\lib.rs:98
  15: alloc::boxed::{{impl}}::call_box<(),closure>
             at C:\projects\rust\src\liballoc\boxed.rs:640
  16: std::sys::imp::thread::{{impl}}::new::thread_start
             at C:\projects\rust\src\libstd\sys\windows\thread.rs:50
  17: BaseThreadInitThunk
```
This one might actually be caused by this PR but I dont understand the message.

e00E:
Aha. These tests that I changed to httpbin.org instead of google.com failed because google detects my german ip adddress and sends a redirect for google.de . Appveyors server probably happen to be in the US as well. I think it still makes sense to move them  to httpbin so that doesnt happen.

steffengy:
No, master passes still today (just reran it to be sure)
The redirect is because google redirects to your countries version of google,
so that test only works for US-google afaik.
Google is used because it's assumed to use a fairly robust/standard https implemenation
and is 100% available.
https://ci.appveyor.com/project/steffengy/schannel-rs/history

The message you're getting (besides the other test failures of this PR)
is SEC_E_ILLEGAL_MESSAGE, which means schannel is getting wrong data.
(likely incomplete)

e00E:
On my machine now all the tests with this PR are working except `test::valid_algorithms`.
Im not sure what to do about the appveyor failures.

src/tls_stream.rs
```diff
@@ -620,10 +620,18 @@ impl<S> TlsStream<S>
...
             self.out_buf.set_position((position + nwritten) as u64);
         }
+        // TODO: Should unused memory in out_buf be freed up here?
+        /*
+        if out > 0 {
+            let position = self.out_buf.position();
+            self.out_buf.set_position(position - out as u64);
+            self.out_buf.get_mut().drain(..out);
+        }
+        */
```
e00E: Yes, No?
Would decrease memory usage and increase allocation count.

src/tls_stream.rs
```diff
@@ -730,18 +738,23 @@ impl<S> TlsStream<S>
...
-                self.out_buf.get_mut().resize(len, 0);
-            }
+        self.out_buf.set_position(0);
+        if self.out_buf.get_ref().len() < len {
+            self.out_buf.get_mut().resize(len, 0);
+        } else {
+            self.out_buf.get_mut().truncate(len); // Does not change capacity
+        }
+        // TODO: Should unused memory in out_buf be freed up here?
+        // self.out_buf.get_mut().shrink_to_fit();
```
e00E: Yes, No?
Would decrease memory usage and increase allocation count.

e00E:
I have fixed the tests originally using google, which I initially moved to httpbin. I think the appveyor servers dont support the right algorithms for httpbin so thats why they failed.
The new tests for google.com should work for everyone as they also accept the moved status code.
I also moved rust stable to `1.1.8.0` so it supports the `?` operator. I hope that was ok, I wasnt sure if you simply forgot to update the rust version or intentionally used an older one.

There are still 3 tests that fail on appveyor: 
```
test test::accept_a_socket ... FAILED
test test::accept_one_byte_at_a_time ... FAILED
test test::split_cert_key ... FAILED
```

These work locally on my machine and I do not know why they fail on appveyor.

src/tls_stream.rs
```diff
@@ -620,10 +620,18 @@ impl<S> TlsStream<S>
         let mut out = 0;
         while self.out_buf.position() as usize != self.out_buf.get_ref().len() {
             let position = self.out_buf.position() as usize;
-            let nwritten = try!(self.stream.write(&self.out_buf.get_ref()[position..]));
+            let nwritten = self.stream.write(&self.out_buf.get_ref()[position..])?;
             out += nwritten;
             self.out_buf.set_position((position + nwritten) as u64);
         }
+        // TODO: Should unused memory in out_buf be freed up here?
```
sfackler: This will not cause any memory to be freed. We don't want to try to shrink out_buf anyway - it'll cap out at the maximum size of a TLS record. Reallocating on every write is pretty terrible perf-wise.

src/tls_stream.rs
sfackler: No.

sfackler:
We intentionally pin to a specific version of rustc - we want to know when we're forcing the minimum supported version forward. Bumping that just so a couple of `try!`s can be3 converted to `?` doesn't seem worth it IMO.

e00E:
I noticed that if I remove the calls to `shutdown` in the failing tests on appveyor work again, so shutdown is probably making some assumption that does not hold with my pr anymore.
Could one of you check that? I dont understand the code well enough and it is hard to debug because it does not fail on my machine.

src/tls_stream.rs
```diff
@@ -730,18 +730,21 @@ impl<S> TlsStream<S>
...
 
-        unsafe {
-            let len = sizes.cbHeader as usize + buf.len() + sizes.cbTrailer as usize;
+        let len = sizes.cbHeader as usize + buf.len() + sizes.cbTrailer as usize;
 
-            if self.out_buf.get_ref().len() < len {
-                self.out_buf.get_mut().resize(len, 0);
-            }
+        self.out_buf.set_position(0);
+        if self.out_buf.get_ref().len() < len {
```
steffengy: This is redundant, resize [implies](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize) truncate for `len() < len`.

src/tls_stream.rs
```diff
@@ -759,9 +762,6 @@ impl<S> TlsStream<S>
 
             match secur32::EncryptMessage(self.context.get_mut(), 0, &mut bufdesc, 0) {
                 winapi::SEC_E_OK => {
-                    let len = bufs[0].cbBuffer + bufs[1].cbBuffer + bufs[2].cbBuffer;
-                    self.out_buf.get_mut().truncate(len as usize);
```
steffengy: Why are you ignoring the actual size of the encrypted data?
I don't see where else the out_buf is restricted to it, this could lead to invalid writes.

e00E: Ah, thanks for finding that. I wrongly thought that the sizes of the buffers passed to `EncryptMessage` would not change. I definitely shouldn't have assumed that.

e00E:
Great, the tests on appveyor pass now. Sorry for the many things you had to review in this PR. The final changes are small. Should I squash the PR to get all the intermediate commits out?

e00E:
I see Github now has that functionality built in for the one merging PRs, so nevermind.

steffengy:
@sfackler 

Since now atleast the tests pass, do you think this approach, 
which does not guarantee a direct-write (requires a call to flush) is viable?
(not violating any guarantees/BC)

Maybe doing a "silent" write_out at the end of `write` would be beneficial
(matching for WouldBlock and ignoring it there to ensure what can be written is written asap).

If not, we'll need to fallback to changing the current approach to only encrypt additional parts on retries,
as discussed.

sfackler:
The behavior with this change behaves a bit like a `BufWriter`, where the behavior before was a bit more like a `TcpStream`. I think I prefer the second behavior - it's a bit weird to require flushes when using a blocking socket. We could solve that by trying to write out the buffer immediately like @steffengy suggested, but it might be simpler to just track the first write buffer size like OpenSSL does.

e00E:
Trying to write out the buffer immediately has as far as I can tell the complexity of:
```rust
match self.write_out() {
    Ok(_) => (), 
    Err(WouldBlock) => (), //ignore
    Err(err) =>  Err,
}
````
Which is not very complex (though `TlsStream` should probably flush on drop then too, like `BufWriter`, which is a few more lines of code). It has the advantage of working correctly no matter how `write` is used.

Tracking the write buffer size is probably in the same realm or less complex but would have surprising results with nonstandard use of `write`. This nonstandard use is probably very unlikely to actually happen.

You two definitely have more experience with this than I do, just restating my argument.

e00E:
...And in either case the behavior should be documented which is what lead to this bug in the first place.

sfackler:
@steffengy I think this seems okay to land as-is, thoughts?

steffengy:
@sfackler 
Sure.
Does this count as breaking change though, since we now require flushes?

steffengy:
@sfackler 
Should we land this first and then take care of that in a separate commit?

sfackler:
👍

PR #38, opened 2017-08-17
Feature/key info

sstecko:
This function adds the ability to gather the Key Usage info in string format from a security context. 

This can be used to filter certs provided to a Tls_Stream in the security context such that non-signature keys are not supplied. I have been experiencing issues where the server requests client certificate validation but the client supplies symmetric (S/MIME) keys, causing the request to crash server-side.

### DIFF
--- a/src/cert_context.rs
+++ b/src/cert_context.rs
@@ -22,6 +22,8 @@ const CRYPT_ACQUIRE_ALLOW_NCRYPT_KEY_FLAG: winapi::DWORD = 0x10000;
 const CRYPT_STRING_BASE64HEADER: winapi::DWORD = 0x0;
 
 lazy_static! {
+    // static ref szOID_KEY_USAGE_RESTRICTION: Vec<u8> =
+    //     winapi::wincrypt::szOID_KEY_USAGE_RESTRICTION.bytes().chain(Some(0)).collect();
     static ref szOID_KEY_USAGE: Vec<u8> =
         winapi::wincrypt::szOID_KEY_USAGE.bytes().chain(Some(0)).collect();
 }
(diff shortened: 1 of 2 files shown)


src/cert_context.rs
```diff
@@ -203,6 +208,35 @@ impl CertContext {
         }
     }
 
+    /// Returns a string that describes the intended key usages of the certificate.
+    pub fn key_usage(&self) -> Option<String> {
+        unsafe {
+            println!("Getting Key usage");
```
sfackler: println

src/cert_context.rs
```diff
@@ -203,6 +208,35 @@ impl CertContext {
...
+            if (*self.0).pCertInfo.is_null() {
```
sfackler: I don't think we need this check.

src/cert_context.rs
```diff
@@ -203,6 +208,35 @@ impl CertContext {
...
+                return None;
+            }
+            let mut keyUsageExtension = crypt32::CertFindExtension(szOID_KEY_USAGE.as_ptr() as winapi::LPSTR, (*(*self.0).pCertInfo).cExtension, (*(*self.0).pCertInfo).rgExtension);
+            if keyUsageExtension.is_null() {
+                return None;
+            } else {
+                let mut len = 500;
```
sfackler: Why 500?

src/cert_context.rs
```diff
@@ -203,6 +208,35 @@ impl CertContext {
...
     }
 
+    /// Returns a string that describes the intended key usages of the certificate.
+    pub fn key_usage(&self) -> Option<String> {
+        unsafe {
+            println!("Getting Key usage");
+            if (*self.0).pCertInfo.is_null() {
+                return None;
+            }
+            let mut keyUsageExtension = crypt32::CertFindExtension(szOID_KEY_USAGE.as_ptr() as winapi::LPSTR, (*(*self.0).pCertInfo).cExtension, (*(*self.0).pCertInfo).rgExtension);
```
sfackler: key_usage_extension

PR #39, opened 2017-08-24
restore behavior to write_out without requiring an explicit flush

steffengy:
cc @sfackler

Not sure this needs to ignore some kinds of errors?
Seems to atleast not break anything immediately, but I'm having
a feeling of missing something here.

Related to #34

### DIFF
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -628,7 +628,6 @@ impl<S> TlsStream<S>
             self.out_buf.set_position((position + nwritten) as u64);
         }
 
-        self.last_write_len = 0;
         Ok(out)
     }
 
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -810,21 +810,17 @@ impl<S> Write for TlsStream<S>
             Some(sizes) => sizes,
             None => return Err(io::Error::from_raw_os_error(winapi::SEC_E_CONTEXT_EXPIRED as i32)),
         };
-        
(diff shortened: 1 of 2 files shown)


sfackler:
We need to ignore WouldBlock errors I believe, since the data's already been successfully buffered and you don't want it to be re-written.

This would still require flushes in nonblocking use though. Maybe the openssl style approach of just tracking the amount of bytes "accepted" by the first read is a better way of solving the initial issue.

steffengy:
So you do favor the "openssl style"?
We still can switch to that, I think we discussed up&down sides thoroughly in #34,
so I won't mention them here in detail again.

I personally think tracking the write buffer length trades a little bit of security
(that shouldn't be hit in any case and will be fine after documenting it) for a small
performance gain (which of course adds up).

src/tls_stream.rs
```diff
@@ -798,22 +802,28 @@ impl<S> MidHandshakeTlsStream<S>
...
 
-        try!(self.encrypt(&buf[..len], &sizes));
-
+        // if we have pending output data, it must have been because a previous
+        // attempt to send this part of the data ran into an error.
+        if self.out_buf.position() == self.out_buf.get_ref().len() as u64 {
+            let left_buf = &buf[self.last_write_len..];
+            self.last_write_len += len;
+            if !left_buf.is_empty() {
+                try!(self.encrypt(left_buf, &sizes));
```
sfackler: I don't think we want to buffer up more data, but wait for what's already there to write out, and then return `Ok(self.last_write_len)`

PR #40, opened 2017-08-24
do not automatically attempt to get a child-certificate chain, fix #37

### LINKED ISSUE #37: Support SCH_CRED_NO_DEFAULT_CREDS flag on AcquireCredentialsHandleA
Hi, 
I'm working on add windows SSL support for MySQL by rust-mysql-simple. And was blocked for the TLSv1.1 Server hello done and Client send cert. Without SCH_CRED_NO_DEFAULT_CREDS flag on AcquireCredentialsHandleA, schannel will pick a cert from current_user/local_machine cert store and send it.  That is not expected and server will response BadHandshake. Must set the SCH_CRED_NO_DEFAULT_CREDS on the AcquireCredentialsHandleA can solve this problem. 

Workaround patch is that in https://github.com/steffengy/schannel-rs/blob/master/src/schannel_cred.rs#L196
Change code to
```rust
cred_data.dwFlags = winapi::SCH_USE_STRONG_CRYPTO | winapi::SCH_CRED_NO_DEFAULT_CREDS;
``` 

Can you provide a way to set the SCH_CRED_NO_DEFAULT_CREDS  flag on the schannel_cred::Builder ?

steffengy:
Unsure if this breaks anything with client-certificates.

Not sure if they even work at this point.

I'd suggest landing this with potentially getting some feedback first
(if we can find someone that uses client certificates)
and then dealing with the fallout.

### DIFF
--- a/src/schannel_cred.rs
+++ b/src/schannel_cred.rs
@@ -193,7 +193,7 @@ impl Builder {
             let mut handle = mem::zeroed();
             let mut cred_data: winapi::SCHANNEL_CRED = mem::zeroed();
             cred_data.dwVersion = winapi::SCHANNEL_CRED_VERSION;
-            cred_data.dwFlags = winapi::SCH_USE_STRONG_CRYPTO;
+            cred_data.dwFlags = winapi::SCH_USE_STRONG_CRYPTO | winapi::SCH_CRED_NO_DEFAULT_CREDS;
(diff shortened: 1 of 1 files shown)


sfackler:
I agree we should land this - if people do want to rely on the automated lookup we can add a flag to opt-back in.

PR #41, opened 2017-08-28
allow retrieving certs as pem or der and their valid use oids

mcgoo:
PEM functionality is used by https://github.com/mcgoo/certstoretopem (which could just as well be an example added to this crate if you like.)

DER functionality will be used for https://github.com/alexcrichton/curl-rust/pull/185

cc @sfackler

### DIFF
--- a/src/cert_context.rs
+++ b/src/cert_context.rs
@@ -301,7 +301,7 @@ impl CertContext {
                         .into_owned(),
                 );
             }
-            Ok(ValidUses::OIDs(oids))
+            Ok(ValidUses::Oids(oids))
         }
     }
 
@@ -565,14 +565,14 @@ impl KeySpec {
     }
 }
 
-/// Valid uses of a Certificate - All, None, or specific OIDs
+/// Valid uses of a Certificate - All, or specific OIDs
 pub enum ValidUses {
     /// Certificate is valid for all uses
     All,
 
     /// Certificate is valid for uses specified. No entries means that the certificate
     /// has no valid uses.
-    OIDs(Vec<String>),
+    Oids(Vec<String>),
 }
 
 #[cfg(test)]
--- a/src/cert_context.rs
+++ b/src/cert_context.rs
@@ -285,10 +285,10 @@ impl CertContext {
             let use_cnt = (*cert_enhkey_usage).cUsageIdentifier;
             if use_cnt == 0 {
                 let last_error = io::Error::last_os_error();
-                return match last_error.raw_os_error() {
-                    Some(winapi::winerror::CRYPT_E_NOT_FOUND) => Ok(ValidUses::All),
-                    Some(0) => Ok(ValidUses::OIDs(Vec::new())),
-                    _ => Err(last_error),
+                match last_error.raw_os_error() {
+                    Some(winapi::winerror::CRYPT_E_NOT_FOUND) => return Ok(ValidUses::All),
+                    Some(0) => (),
+                    _ => return Err(last_error),
                 };
             }
 
--- a/src/cert_context.rs
+++ b/src/cert_context.rs
@@ -287,7 +287,7 @@ impl CertContext {
                 let last_error = io::Error::last_os_error();
                 return match last_error.raw_os_error() {
                     Some(winapi::winerror::CRYPT_E_NOT_FOUND) => Ok(ValidUses::All),
-                    Some(0) => Ok(ValidUses::None),
+                    Some(0) => Ok(ValidUses::OIDs(Vec::new())),
                     _ => Err(last_error),
                 };
             }
@@ -570,10 +570,8 @@ pub enum ValidUses {
     /// Certificate is valid for all uses
     All,
 
-    /// Certificate is not valid for any use
-    None,
-
-    /// Certificate is valid for uses specified
+    /// Certificate is valid for uses specified. No entries means that the certificate
+    /// has no valid uses.
     OIDs(Vec<String>),
 }
 
--- a/src/cert_context.rs
+++ b/src/cert_context.rs
@@ -1,6 +1,6 @@
 //! Bindings to winapi's `PCCERT_CONTEXT` APIs.
 
-use std::ffi::OsString;
+use std::ffi::{CStr, OsString};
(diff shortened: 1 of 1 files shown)


src/cert_context.rs
```diff
@@ -119,6 +124,43 @@ impl CertContext {
...
+                (*self.0).cbCertEncoded,
+                0 as winapi::DWORD,
+                buf.as_mut_ptr() as *mut i8,
+                &mut len,
+            );
+            if ok != winapi::TRUE {
+                return Err(io::Error::last_os_error());
+            }
+
+            buf.pop();
```
sfackler: Is this a newline or something?

sfackler: Looks like you can pass `CRYPT_STRING_NOCRLF` to disable that.

src/cert_context.rs
```diff
@@ -119,6 +124,43 @@ impl CertContext {
...
     }
 
+    /// Get certificate as PEM-formatted X509 certificate.
+    pub fn to_pem(&self) -> io::Result<String> {
+        unsafe {
+            let mut len = 0;
+            let ok = crypt32::CryptBinaryToStringA(
+                (*self.0).pbCertEncoded,
+                (*self.0).cbCertEncoded,
+                0 as winapi::DWORD,
```
sfackler: Could you explicitly pass `CRYPT_STRING_BASE64HEADER` here? It's equivalent but a bit more self documenting IMO.

src/cert_context.rs
```diff
@@ -228,7 +270,48 @@ impl CertContext {
         }
     }
 
-    fn get_encoded_bytes(&self) -> &[u8] {
+    /// Returns the valid uses for this certificate
+    pub fn valid_uses(&self) -> io::Result<ValidUses> {
+        unsafe {
+            let mut buf_len = 0;
+            let ok = crypt32::CertGetEnhancedKeyUsage(self.0, 0, ptr::null_mut(), &mut buf_len);
```
sfackler: Seems like you should be able to just stack allocate a `CERT_ENHKEY_USAGE` in and avoid the double call and vector.

src/cert_context.rs
```diff
@@ -488,6 +571,18 @@ impl KeySpec {
     }
 }
 
+/// Valid uses of a Certificate - All, None, or specific OIDs
+pub enum ValidUses {
+    /// Certificate is valid for all uses
+    All,
+
+    /// Certificate is not valid for any use
+    None,
```
sfackler: This is equivalent to `OIDs` with an empty vector, right?

src/cert_context.rs
mcgoo: It's a trailing \0 - I'll see if it looks cleaner with CStr() if you like?

src/cert_context.rs
mcgoo: The docs don't really call it out, but I verified that the struct has a variable sized trailer.

mcgoo:
@sfackler This should be ready for review again. Thanks.

src/cert_context.rs
sfackler: Ah, gotcha. I think it's fine with the pop if you stick a comment in there saying that's what's going on.

src/cert_context.rs
```diff
@@ -488,6 +565,16 @@ impl KeySpec {
...
 }
 
+/// Valid uses of a Certificate - All, None, or specific OIDs
+pub enum ValidUses {
+    /// Certificate is valid for all uses
+    All,
+
+    /// Certificate is valid for uses specified. No entries means that the certificate
+    /// has no valid uses.
+    OIDs(Vec<String>),
```
sfackler: This should be called `Oids` I believe, to match up with abbreviation capitalization conventions.

src/cert_context.rs
```diff
@@ -228,7 +264,48 @@ impl CertContext {
...
+            if ok != winapi::TRUE {
+                return Err(io::Error::last_os_error());
+            }
+
+            let use_cnt = (*cert_enhkey_usage).cUsageIdentifier;
+            if use_cnt == 0 {
+                let last_error = io::Error::last_os_error();
+                return match last_error.raw_os_error() {
+                    Some(winapi::winerror::CRYPT_E_NOT_FOUND) => Ok(ValidUses::All),
+                    Some(0) => Ok(ValidUses::OIDs(Vec::new())),
```
sfackler: This case can just fall through to below, right?

src/cert_context.rs
```diff
@@ -488,6 +565,16 @@ impl KeySpec {
     }
 }
 
+/// Valid uses of a Certificate - All, None, or specific OIDs
```
sfackler: Needs to be updated for None removal.

src/cert_context.rs
```diff
@@ -228,7 +264,48 @@ impl CertContext {
         }
     }
 
-    fn get_encoded_bytes(&self) -> &[u8] {
+    /// Returns the valid uses for this certificate
+    pub fn valid_uses(&self) -> io::Result<ValidUses> {
```
sfackler: It might make sense to return an `Option<Vec<String>>`?

mcgoo: I'm not sure that works. There are 4 return types - API failure, No uses valid, All uses valid, and a list of valid uses. I don't want to make `None` mean "API failure or all uses valid." (I found the underlying API very confusing to understand because `.cUsageIdentifier==0` means one of "none" or "all" depending on the value of `GetLastError()`.)

I suppose it could return Result<Option<Vec<String>>> and document that Ok(None) means "all" but I think how it is now is more obvious than that.

How ever you like of course... let me know how you want it.

sfackler: Ok yeah, I guess a custom enum is a bit more clear.

sfackler:
@steffengy LGTY?

steffengy:
@sfackler 
Yep, LGTM.

I think we should release 0.1.8 after landing this.
Also it doesn't seem that anything else will land soon and we have
quite a bit changes that have only landed in master so far.
What do you think?

sfackler:
Sounds reasonable!

PR #43, opened 2017-11-17
Fix handling of renegotiation requests

### LINKED ISSUE #42: TlsStream blocking indefinitely when using client cert
When trying to connect and read from an IIS application that requires a client cert, `<TlsStream as BufRead>::fill_buf` appears to be blocking indefinitely, waiting for data that never appears.

I've cloned the repo and added some simple logging, which indicates where the blocking occurs...

```
DEBUG:schannel::tls_stream: Begin TlsStream::initialize
DEBUG:schannel::tls_stream: Writing from 0
DEBUG:tls_test: Reading response
DEBUG:schannel::tls_stream: TlsStream empty. Filling...
DEBUG:schannel::tls_stream: Initializing TlsStream
DEBUG:schannel::tls_stream: Begin TlsStream::initialize
DEBUG:schannel::tls_stream: TlsStream needs read
DEBUG:schannel::tls_stream: begin read_in
DEBUG:schannel::tls_stream:     Reading 2048 bytes...
DEBUG:schannel::tls_stream:     Read 37 bytes
DEBUG:schannel::tls_stream: Decrypting
DEBUG:schannel::tls_stream: TlsStream empty. Filling...
DEBUG:schannel::tls_stream: Initializing TlsStream
DEBUG:schannel::tls_stream: Begin TlsStream::initialize
DEBUG:schannel::tls_stream:     state is State::Inititializing
DEBUG:schannel::tls_stream:     Writing
DEBUG:schannel::tls_stream:     Validating
DEBUG:schannel::tls_stream:     Setting validated to true
DEBUG:schannel::tls_stream:     needs 1 more bytes
DEBUG:schannel::tls_stream: begin read_in
DEBUG:schannel::tls_stream:     Reading 2048 bytes...
<No further data is received at this point>
```

My test application uses `schannel-rs` via `rust-native-tls` and looks like the following (*note: I've added a timeout on the `TcpStream` to prevent blocking in this instance*)...

```rust
    static HTTP_REQ: &'static [u8] = b"GET /broker.asmx HTTP/1.1\r\n\
                                  Host: [HOSTNAME]\r\n\
                                  \r\n";

    let mut builder = TlsConnector::builder()
        .unwrap();

    builder
        .add_root_certificate(load_server_cert())
        .unwrap()
        .identity(load_client_cert())
        .unwrap()
        ;

    let connector = builder.build()
        .unwrap();

    let stream = TcpStream::connect("[IP & PORT]").unwrap();
    stream.set_read_timeout(Some(Duration::from_secs(5))).unwrap();

    debug!("Connecting TLS");
    let mut stream = connector.connect("[HOSTNAME]", stream).unwrap();

    debug!("Writing request...\r\n{}", str::from_utf8(HTTP_REQ).unwrap());
    stream.write_all(HTTP_REQ)
        .unwrap();

    let mut response = vec![0_u8; 256];
    debug!("Reading response");
    loop {
        match stream.read(&mut response) {
            Ok(n) => print!("{}", String::from_utf8_lossy(&response[..n])),
            Err(e) => {
                println!("{}", e);
                break;
            }
        }
    }
```

I'm not familiar with the SChannel's operation so forgive the speculation, but it looks like maybe SChannel is waiting on a write to occur before it can read more data.

I'm happy to take direction on how to diagnose further.

sfackler:
If a renegotiation's been requested, we're responsible for the next
message, so we definitely don't need to read.

Closes #42

@gmbeard can you try out this branch to make sure it works as you'd expect?

### DIFF
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -722,7 +722,7 @@ impl<S> TlsStream<S>
                         self.enc_in.position() as usize
                     };
                     self.consume_enc_in(nread);
-                    self.needs_read = (self.enc_in.position() == 0) as usize;
+                    self.needs_read = 0;
                     Ok(false)
                 }
                 e => Err(io::Error::from_raw_os_error(e as i32)),


gmbeard:
You beat me to it! I was gonna tidy up some commits and send a PR myself. 

I shall indeed try this out later on today. Great work. Thanks for the effort

gmbeard:
@sfackler I'm still getting blocked reads with this commit. I think the problem lies in the [block of code at L704:710](https://github.com/steffengy/schannel-rs/pull/43/files#diff-de9f1ab4b207a1a4254c21176c722b8cR704). I'm not quite sure what's going on here but branching this with...

```rust
if r == winapi::SEC_E_OK {
    ...
}
```

... seems to fix it. I'm guessing it's discarding encrypted bytes at the end of `self.enc_in` when it shouldn't be.

steffengy:
Since this doesn't seem as trivial as expected to get right, we might want a testcase for this?

If I understand [this](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379413(v=vs.85).aspx
) correctly it should be feasible to do a client<->server test 
and force renegotiation to test it works as intended?

sfackler:
Yeah, I'll write up a testcase tonight.

gmbeard:
@sfackler, anything I can help with?

sfackler:
Ok, I think this should be the actual fix. I've tested manually with openssl s_server. I was unfortunately having some trouble correctly triggering a renegotiation though schannel.

gmbeard:
I can confirm that this latest commit resolves my issue 👍

PR #44, opened 2017-12-18
migrate to winapi 0.3

steffengy:
this is just an upgrade of internals and adds no new behavior (hopefully)


sfackler:
I think we may now be able to get rid of the libncrypt archives as well.

steffengy:
@sfackler 
Yep finally, took care of that aswell.
I also tried to find all FIXME/TODOS related to this, hopefully I got most of them.

Time improvement is clearly noticable: An appveyor test run took about 7 minutes before,
now we're down to about 5 minutes.

I used `use winapi::shared::minwindef as winapi;` to keep the diff somewhat better readable
and grew to actually not hating it that way, so let me know during your review if you
think it's better to get rid of that :)

sfackler:
The module paths in winapi match up to include paths of the C header, right? The current diff seems reasonable to me.

steffengy:
> The module paths in winapi match up to include paths of the C header, right? The current diff seems reasonable to me.

That was the plan as far as I know and also seemed to be the case where I checked.
I quite like how much faster it's locally, not sure if that's due to a new rust version 
but it feels at least twice as fast.

Do we have any exposed surface for winapi?
(if so, we'd have to consider this a breaking change and make sure to bump to 0.2 after merging this)

sfackler:
I don't think we expose winapi anywhere so it shouldn't require an 0.2.

steffengy:
yeah, and if there was an winapi type in some public API, I'd likely consider that the issue.
I also didn't spot anything while looking through the docs.

PR #45, opened 2018-01-25
Fix/check invalid intermediate certs

sstecko:
This fixes a bug where we would fail to construct a full cert-chain in cases where an intermediate cert is invalid but is signed by a valid parent cert. 

This brings us more in line with chromium https://cs.chromium.org/chromium/src/net/cert/cert_verify_proc_win.cc?rcl=1fe0376cc89cf15de73cfe8fc10fca6c515484c6&l=853


sstecko:
Wooooooops. Wrong set of commits

PR #46, opened 2018-02-24
support use of intermediate certificate authorities

mdonoughe:
It looks like in order to support certificate chains longer than two certificates the `hAdditionalStore` parameter to `CertGetCertificateChain` needs to contain the certificate chain provided by the remote side, available through the `hCertStore` member of the certificate context (remarks on [QueryContextAttributes (General) function](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379326(v=vs.85).aspx)). See [Chromium cert_verify_proc_win.cc](https://chromium.googlesource.com/chromium/src/net/+/6a2d97ad76e85f195ec1edaf9dabbf08b9a8ab1d/cert/cert_verify_proc_win.cc#996).

I'm not sure I've done this correctly because I haven't used the schannel API directly before. This seems to work when the old code didn't.

Was there a reason for iteration of certificates in a store to require mutability of the store? I changed that because I was having trouble merging the local store with the remote store otherwise and it didn't seem right.

Is there a better way to test this than connecting to lh3.googleusercontent.com? That's how I found the problem, but there's no guarantee they won't change their certificates in a way that invalidates the test or even discontinue that domain altogether. I just copied the test above it that connects to google.com.

### DIFF
--- a/src/cert_context.rs
+++ b/src/cert_context.rs
@@ -14,6 +14,7 @@ use winapi::um::wincrypt;
 use Inner;
 use ncrypt_key::NcryptKey;
 use crypt_prov::{CryptProv, ProviderType};
+use cert_store::CertStore;
 
 /// A supported hashing algorithm
 pub struct HashAlgorithm(winapi::DWORD, usize);
@@ -296,6 +297,19 @@ impl CertContext {
                 );
             }
             Ok(ValidUses::Oids(oids))
+        }
+    }
+
+    /// For a remote certificate, returns a certificate store containing any intermediate
+    /// certificates provided by the remote sender.
+    pub fn cert_store(&self) -> Option<CertStore> {
+        unsafe {
+            let chain = (*self.0).hCertStore;
+            if chain.is_null() {
+                None
+            } else {
+                Some(CertStore::from_inner(wincrypt::CertDuplicateStore(chain)))
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


mdonoughe:
It looks like the CI fails because `?` was not stable until Rust 1.13 2016-11-10. I don't know if compatibility with old versions of Rust is important here.

steffengy:
> It looks like the CI fails because ? was not stable until Rust 1.13 2016-11-10. I don't know if compatibility with old versions of Rust is important here.

It's more of an issue of consistency, since the rest of the code still uses `try`.
I personally don't see a reason to break that backwards compatibility, but
that'd be another issue/discussion anyways, so let's stick to `try` for now.

The approach generally looks fine to me at a first glance.

steffengy:
> Changing &mut self to &self is technically a back compat break. I think it'd probably be just fine to change, but we could also just make certs2 or whatever and deprecate the old method.

Yeah sure technically, but relaxing from `&mut self` to `&self` contrary to going from `&self` to `&mut self` shouldn't break any existing code, so I don't think this warrants adding a second method.
Or is there something I'm not taking into account here?

sfackler:
You could observe it if you were doing something like this:

```rust
let f: fn(&mut CertContext) -> Certs = CertContext::certs;
```

It's pretty artificial example though.

fxb:
It seems like this fixes an issue I've encountered when using sccache.

See: https://github.com/mozilla/sccache/issues/228

Any plans to bump and release a new version?

steffengy:
@fxb Done, released as v0.1.11.

PR #48, opened 2018-04-17
Allow hostname verification and SNI to be separately configured

### LINKED ISSUE #47: Allow hostname verification and SNI to be separately configured
They're both enabled when a domain is set, but we should let you e.g. use SNI but not verify the hostname or vice versa.

sfackler:
Closes #47 

r? @steffengy

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -184,6 +184,19 @@ fn wrong_host_cert() {
 }
 
 #[test]
+fn wrong_host_cert_ignored() {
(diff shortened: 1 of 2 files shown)


steffengy:
@sfackler 
are there any soon(ish) plans for other changes on your side or should I directly cut 0.1.12?

sfackler:
Nope - we can cut 0.1.12 now!

PR #51, opened 2018-06-03
Add TlsStream::is_server

sfackler:
This is the equivalent to OpenSSL's SSL_is_server.

### DIFF
--- a/src/security_context.rs
+++ b/src/security_context.rs
@@ -1,4 +1,5 @@
 use winapi::shared::{sspi, winerror};
+use winapi::shared::minwindef::ULONG;
 use winapi::um::{minschannel};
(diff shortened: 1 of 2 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
steffengy: Is this file intentionally committed?

sfackler: Oops, fixed!

sfackler:
For reference, I'm using these over here: https://github.com/sfackler/rust-native-tls/pull/97/files#diff-f3a4e813fbd2200f75435edfe7dcba3dR297

PR #52, opened 2018-09-25
enable sspi feature in winapi (fixes build with winapi 0.3.6)

mcgoo:
`schannel` does not build with the current `winapi` (0.3.6) because the `sspi` feature is no longer being implicitly enabled by the set of features that `schannel` enables. I think this is the relevant commit: https://github.com/retep998/winapi-rs/commit/def700cbe3044d22d044cade4202ca194c6a6623


mcgoo:
I'm not sure why the tests are failing. For what it's worth, the failures look the same on my machine when building with `winapi` 0.3.5 and 0.3.6.

@sfackler since you released the last crate.

steffengy:
@mcgoo tests are fixed - just an expired test certificate.

@sfackler 
it seems like winapi uses `associated_consts`, any objections to bumping to atleast rust 1.20?
(we really don't have a choice, if winapi doesn't work with anything lower)

retep998:
That's `lazy_static`'s fault, not `winapi`'s fault!

steffengy:
@retep998 oops, thanks for pointing that out, my bad.

sfackler:
It seems reasonable to either explicitly downgrade lazy_static in the 1.10 build via `cargo update -p lazy-static --precise 1.0.2` or just bump the Rust version to whatever's needed for associated constants.

steffengy:
Thanks, I'll look into cutting a release soon, since this might result in failed builds for some people.

PR #53, opened 2019-02-06
return subject public key info for cert that can be used for pinning

SergejJurecko:


SergejJurecko:
I have finished messing with it. It works well in my pinning tests.

retep998:
Looks like the minimum Rust version needs to be bumped *again* due to `lazy_static`.

SergejJurecko:
Guess I spoke too soon. I bumped version to minimum used in lazy_static and fixed my own code to be compatible with old rustc

steffengy:
@SergejJurecko 
Looks quite good so far, could you change it to not use CRYPT_ENCODE_ALLOC_FLAG, as we do in similar cases in that file?

(As in the first call to CryptEncodeObjectEx returns the needed size, allocating the Vec in rust and passing it to the second call of CryptEncodeObjectEx)

SergejJurecko:
Done

src/cert_context.rs
```diff
@@ -86,6 +86,34 @@ impl CertContext {
...
+        unsafe {
+            let mut len:u32 = 0;
+            wincrypt::CryptEncodeObjectEx(wincrypt::X509_ASN_ENCODING,
+                wincrypt::CERT_INFO_SUBJECT_PUBLIC_KEY_INFO_FLAG as *const u32 as *const _,
+                &(*(*self.0).pCertInfo).SubjectPublicKeyInfo as *const wincrypt::CERT_PUBLIC_KEY_INFO as _,
+                0,
+                ptr::null_mut(),
+                ptr::null_mut(),
+                &mut len as *mut _
+                );
```
steffengy: identation is off here

src/cert_context.rs
```diff
@@ -86,6 +86,34 @@ impl CertContext {
...
+            if len > 0 {
+                let mut buf = vec![0; len as usize];
+                wincrypt::CryptEncodeObjectEx(wincrypt::X509_ASN_ENCODING,
+                    wincrypt::CERT_INFO_SUBJECT_PUBLIC_KEY_INFO_FLAG as *const u32 as *const _,
+                    &(*(*self.0).pCertInfo).SubjectPublicKeyInfo as *const wincrypt::CERT_PUBLIC_KEY_INFO as _,
+                    0,
+                    ptr::null_mut(),
+                    buf.as_mut_ptr() as _,
+                    &mut len as *mut _,
+                    );
```
steffengy: and here

src/cert_context.rs
```diff
@@ -86,6 +86,34 @@ impl CertContext {
...
+                wincrypt::CERT_INFO_SUBJECT_PUBLIC_KEY_INFO_FLAG as *const u32 as *const _,
+                &(*(*self.0).pCertInfo).SubjectPublicKeyInfo as *const wincrypt::CERT_PUBLIC_KEY_INFO as _,
+                0,
+                ptr::null_mut(),
+                ptr::null_mut(),
+                &mut len as *mut _
+                );
+            if len > 0 {
+                let mut buf = vec![0; len as usize];
+                wincrypt::CryptEncodeObjectEx(wincrypt::X509_ASN_ENCODING,
```
steffengy: Also you should check the return value for both calls, as in other places:

```rust
let ok = wincrypt::CryptEncodeObjectEx(...);
if ok != winapi::TRUE {
                return Err(io::Error::last_os_error());
}
```

SergejJurecko:
Done

steffengy:
@sfackler so far fine from my side, any objections from yours?

steffengy:
Released as v0.1.15.

PR #56, opened 2019-08-29
Add PKCS8 Key Import Support

Goirad:
This PR is an initial implementation for importing a key into `ImportOptions`, and is required for [this PR](https://github.com/sfackler/rust-native-tls/issues/27) in `native-tls`.

### DIFF
--- a/src/crypt_prov.rs
+++ b/src/crypt_prov.rs
@@ -1,6 +1,7 @@
 //! CryptoAPI key providers.
 use std::io;
 use std::ptr;
+use std::slice;
 use winapi::shared::minwindef as winapi;
 use winapi::um::winbase;
 use winapi::um::wincrypt;
@@ -204,7 +205,7 @@ impl<'a> ImportOptions<'a> {
         }
     }
 
-    /// Imports a DER-encoded PKCS8 pricate key.
+    /// Imports a DER-encoded PKCS8 private key.
     pub fn import_pkcs8(&mut self, der: &[u8]) -> io::Result<CryptKey> {
         unsafe {
             assert!(der.len() <= winapi::DWORD::max_value() as usize);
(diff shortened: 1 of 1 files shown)


sfackler:
Probably worth adding a simple test as well.

Goirad:
I added a test and moved up the free, though it looks like CI is failing for unrelated reasons. Any ideas as to what might be going on?

steffengy:
@Goirad Please rebase against master, CI is fixed and all green again.

Goirad:
@steffengy rebased, would love any feedback or suggestions

steffengy:
lgtm

src/crypt_prov.rs
```diff
@@ -203,11 +203,62 @@ impl<'a> ImportOptions<'a> {
...
+                                                    wincrypt::CRYPT_DECODE_ALLOC_FLAG,
+                                                    ptr::null_mut(),
+                                                    &mut buf as *mut _ as winapi::LPVOID,
+                                                    &mut len);
+            if res == winapi::FALSE {
+                return Err(io::Error::last_os_error());
+            }
+            let pkey: wincrypt::CRYPT_PRIVATE_KEY_INFO = *buf;
+
+            // Decode pkey's internal der blob again into the desired DSS V3 Private Key BLOB
```
jethrogb: Can this be implemented by calling `self.import`?

Goirad: Yes it can, done

Goirad:
@jethrogb looking [here](https://docs.microsoft.com/en-us/windows/win32/seccrypto/constants-for-cryptencodeobject-and-cryptdecodeobject) it looks like there is functionality for decoding ECC keys, probably in a way very similar to the new PKCS8 function

steffengy:
As we currently do not support importing ECC keys, I think if there're no further objections we're fine landing this as is for now.

Goirad:
None from me

PR #59, opened 2019-09-11
Update for 2018 edition

Goirad:
Makes all the changes for 2018 edition

### DIFF
--- a/src/key_handle.rs
+++ b/src/key_handle.rs
@@ -1,5 +1,3 @@
 //! Deprecated.
-#![deprecated(note = "use cert_context::PrivateKey", since = "0.1.5")]
-
 #[deprecated(note = "use cert_context::PrivateKey", since = "0.1.5")]
-pub use cert_context::PrivateKey as KeyHandle;
+pub use crate::cert_context::PrivateKey as KeyHandle;
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,7 @@ environment:
   - TARGET: x86_64-pc-windows-gnu
     VERSION: nightly
   - TARGET: i686-pc-windows-gnu
-    VERSION: 1.27.1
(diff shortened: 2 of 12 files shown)


Goirad:
I'm realizing now that one of the tests is against compiler version 1.27, how come?

steffengy:
Well, we generally try to stay compatible with the oldest possible version.
Due to dependencies 1.27 is the lowest we can go.

We can go for 1.31 (which is I think all that is required for edition 2018).
I'll probably delay a crates.io release for a while though.

Goirad:
>  I'll probably delay a crates.io release for a while though.
In general, or if we go for 1.31?

steffengy:
In general, but a bit more than usual with 1.31.
Not more than 2 weeks, just giving people a tidbit of time to notice ;)
But you might want to get further changes in anyways in that time window?

Goirad:
Yea that's fair, I probably will have to do more work for [this](https://github.com/sfackler/rust-native-tls/issues/132) anyways.

src/key_handle.rs
```diff
@@ -1,5 +1,3 @@
 //! Deprecated.
-#![deprecated(note = "use cert_context::PrivateKey", since = "0.1.5")]
-
 #[deprecated(note = "use cert_context::PrivateKey", since = "0.1.5")]
-pub use cert_context::PrivateKey as KeyHandle;
+pub use crate::cert_context::PrivateKey as KeyHandlendle;
```
steffengy: `KeyHandlendle` looks like a typo

Goirad: Fixed, but that is bizarre, since that would have been `cargo fix`

steffengy: Usually when I have that it's IDE/editor lag, but yeah very strange ^^

steffengy:
Thanks!

PR #60, opened 2019-09-23
Added Slightly More Docs Regarding Common Trust Locations

bmarcaur:
Wanted to make this a bit more type safe using an `enum` but after reading the MSDN docs it seems that the `pvPara` param is pretty "varied" depending on the context of other params. So as a small improvement just added the 2 other common locations to the docs.

### DIFF
--- a/src/cert_store.rs
+++ b/src/cert_store.rs
@@ -102,7 +102,8 @@ pub enum CertAdd {
 impl CertStore {
     /// Opens up the specified key store within the context of the current user.
     ///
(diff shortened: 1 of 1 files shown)


bmarcaur:
Any objections @steffengy ?

steffengy:
No, sorry for the long turnaround. 
This being a minor doc change made this not a high priority for me to review back at the time.

bmarcaur:
Of course! No worries. Thank you for the library and the merge.

PR #64, opened 2020-01-20
add method to import PEM formatted PKCS8 keys

Goirad:
I have added a method to import PEM formatted keys. Ideally, the other method would be changed to `import_pkcs8_der`, but that would require a major version bump. 

Also of note, windows does not validate the PEM headers, and only checks that the header matches the footer. It is unclear to me whether this crate is the place to add that logic, or whether it would be sufficient to add rudimentary checks on the input string in the `import_pkcs8_pem` method.

Another alternative would be add something like a `parse_pkcs8_pem` method and leave it to the user of this library to call that first before then calling `import_pkcs8` on the result. I would be happy to use that approach instead and include the header checking there if that is a preferred structure.


src/crypt_prov.rs
```diff
@@ -283,4 +321,70 @@ mod test {
...
+            .unwrap();
+        assert!(context.import()
+            .import_pkcs8_pem(key)
+            .is_err());
+    }
+
+    #[test]
+    // FIXME this test should fail, but it does not
+    // It appears windows does not check that the header is correct,
+    // only that it matches the footer
```
jethrogb: Not sure if it's useful to have these tests here

src/crypt_prov.rs
```diff
@@ -233,6 +234,43 @@ impl<'a> ImportOptions<'a> {
...
+    pub fn import_pkcs8_pem(&mut self, pem: &[u8]) -> io::Result<CryptKey> {
+        unsafe {
+            assert!(pem.len() <= winapi::DWORD::max_value() as usize);
+
+            // Decode the pem wrapper before passing it to import_pkcs8
+            // Call once first to figure out the necessary buffer size
+            let mut len = 0;
+            let res = wincrypt::CryptStringToBinaryA(pem.as_ptr() as ntdef::LPCSTR,
+                                                    pem.len() as winapi::DWORD,
+                                                    wincrypt::CRYPT_STRING_BASE64_ANY,
```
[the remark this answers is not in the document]
steffengy: I do not think so, parsing should work for PEM files with & without header.

src/crypt_prov.rs
```diff
@@ -233,6 +234,43 @@ impl<'a> ImportOptions<'a> {
             res
         }
     }
+
+    /// Imports a PEM-encoded PKCS8 private key.
+    /// This functions decodes PEM blocks with or without "BEGIN PUBLIC KEY"
+    /// and "END PUBLIC KEY" headers
```
steffengy: should be "BEGIN/END PRIVATE KEY" right?

steffengy:
I think windows only checks if the header begins with a "BEGIN".
I'd think I'd prefer the `import_pkcs8_pem` approach and to fullfill the contract of the name some rudimentary checks (if it starts with "----BEGIN" it has to be "-----BEGIN PRIVATE KEY-----" and has to end with "-----END PRIVATE KEY-----")
What do you think?

src/crypt_prov.rs
Goirad: yes it should

Goirad:
Sounds good to me, what do you think of renaming `import_pkcs8` to `import_pkcs8_der`?

jethrogb:
What happens if you try to pass in a PKCS1 key?

Goirad:
Turns out it rejects PKCS1 keys. I've added a test to that effect. Also it seems tests are failing for unrelated reasons.

steffengy:
> Sounds good to me, what do you think of renaming `import_pkcs8` to `import_pkcs8_der`?

I think `import_pkcs8_pem` is better for consistency reasons. In all places I found we currently have pem explicitly as suffix and omit DER.

Goirad:
Is there anything else that you want me to change for this PR?

steffengy:
@Goirad It's fine from my side, maybe @sfackler wants some changes before this fits into how you intend to use it in native-tls - atleast that's what I gathered.
(Referencing https://github.com/sfackler/rust-native-tls/pull/147)

Goirad:
@steffengy Can we get a release?

steffengy:
@Goirad Sure, done.

PR #67, opened 2020-03-04
Allow the same SchannelCred to be used for multiple connections to support TLS session ticket resumption

### LINKED ISSUE #66: Interface changes to support RFC 5077 TLS session tickets
It looks like SChannel natively advertises support for the `SessionTicket` extension in its `ClientHello`, but scopes its stored tickets in some way to the `CredHandle` used to build the session.

As currently written, `tls_stream::Builder` consumes the given `SchannelCred`, so it's not possible to use this crate in a way that allows for RFC 5077. https://github.com/steffengy/schannel-rs/blob/3becee06f5117108ce342f230abed6239417e07b/src/tls_stream.rs#L119-L122

Would it be possible to update the interface to allow `SchannelCred` to be shared/refcounted? At a glance through the `curl`'s usage of `schannel`, it looks like the Windows APIs that use `CredHandle` objects are most likely thread-safe.

Ref: https://github.com/curl/curl/blob/0e06c1637b30800d41636dcd02f5becf3f6664c1/lib/vtls/schannel.c#L794

cyang1:
Adds an `Arc` within the `SchannelCred`, and slightly reworks the logic that uses it as a result. Not sure what to do about `get_mut()` at this point.

Added tests for the behavior by fetching the `SECPKG_ATTR_SESSION_INFO` attribute off the security context, which has information about whether the session was resumed.

Fixes #66.

### DIFF
--- a/src/tls_stream.rs
+++ b/src/tls_stream.rs
@@ -10,7 +10,7 @@ use std::slice;
 use std::sync::Arc;
 use winapi::shared::minwindef as winapi;
 use winapi::shared::{ntdef, sspi, winerror};
-use winapi::um::{self, wincrypt};
+use winapi::um::{self, schannel, wincrypt};
 
 use crate::{INIT_REQUESTS, ACCEPT_REQUESTS, Inner, secbuf, secbuf_desc};
 use crate::cert_chain::{CertChain, CertChainContext};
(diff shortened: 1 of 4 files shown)


src/tls_stream.rs
```diff
@@ -352,6 +352,20 @@ impl<S> TlsStream<S>
         self.context.remote_cert()
     }
 
+    /// Returns whether or not the session was resumed.
+    ///
+    /// # Remarks
+    /// This is only valid if the TLS handshake is completed, and will return None if called
+    /// before this point.
+    pub fn session_resumed(&self) -> io::Result<Option<bool>> {
+        if let State::Streaming { .. } = self.state {
```
steffengy: Is there a reason for this check? 
E.g. would calling `self.context.session_info()` similarily to `remote_cert` fail, if the handshake wasn't completed yet?

cyang1: How is `remote_cert()` expected to fail if the handshake wasn't completed?

It looks like `self.context.session_info()` will fail with `SEC_E_INVALID_HANDLE` before the handshake completes.

I'm okay with changing this to just `pub fn session_resumed(&self) -> io::Result<bool>` if the above matches existing expected behavior.

cyang1: It looks like `remote_cert()` fails with the same error. I'll remove the check.

steffengy:
@cyang1 LGTM. and especially thanks for the tests 👍 .
If you want, you can remove the `Inner` implementation for SchannelCred and move the relevant methods we use there to SchannelCred directly, so nobody can call the unimplemented `get_mut` method.

PR #76, opened 2022-05-04
Switched from winapi to windows-rs

### LINKED ISSUE #75: Switch to official Microsoft-owned windows-rs crate
The winapi-rs seems to be unmaintained and in general deprecated in favor of the official Microsoft Rust project at https://github.com/microsoft/windows-rs which in turn generates it's APIs from the https://github.com/microsoft/win32metadata

I am wondering if it makes sense to switch to the official crate and if such a change would be accepted by maintainers.
I am willing to submit a PR for this.

ancwrd1:
This PR resolves #75. Important bullet points:
* I have used `windows-sys` crate which is roughly equivalent to winapi. The higher-level `windows` crate doesn't currently work well with the functions returning HRESULT and suffers from occasional breaking changes in the APIs
* The MSRV is 1.40 due to added `non_exhaustive` pattern which replaces the manual one on several enums
* The code was formatted with rustfmt and updated to 2018 edition
* Fixed a few minor clippy warnings
* Tested with both -msvc and -gnu targets
* Mostly translates one to one with a few tweaks around mutability and types

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,10 +8,17 @@ documentation = "https://docs.rs/schannel/0.1.19/schannel/"
 repository = "https://github.com/steffengy/schannel-rs"
 readme = "README.md"
 keywords = ["windows", "schannel", "tls", "ssl", "https"]
+edition = "2018"
 
 [package.metadata.docs.rs]
 default-target = "x86_64-pc-windows-msvc"
 
 [dependencies]
 lazy_static = "1.0"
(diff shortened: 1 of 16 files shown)


steffengy:
Fixed appveyor tests and checked that this builds through CI given with 1.40 as min version: https://ci.appveyor.com/project/steffengy/schannel-rs/builds/43467207

Build times seem the same, so no issue on that front.
Need to still take a look at the code and check some compatiblity things.

steffengy:
Thanks

PR #78, opened 2022-07-28
Remove dependency on `lazy_static`

Expyron:
Uses a macro to null-terminate static strings at compile time.

This raises the MSRV of this crate to 1.46 due to the need for a loop in const-eval.


Expyron:
This requires #79 for the CI tests to pass.

PR #82, opened 2022-10-20
Update windows-sys to 0.42

lazka:
https://github.com/microsoft/windows-rs/releases/tag/0.42.0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,10 +14,10 @@ edition = "2018"
 default-target = "x86_64-pc-windows-msvc"
 
 [dependencies]
-windows-sys = { version = "0.36", features = [
+windows-sys = { version = "0.42", features = [
     "Win32_Foundation", "Win32_Security_Cryptography",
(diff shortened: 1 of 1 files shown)


messense:
Ping, the ecosystem is moving to windows-sys 0.42.0, `schannel` is the only one depends on 0.36 in my dependency tree right now. Please merge this and publish a new version, thanks a lot!

messense:
I suppose this is blocked by the MSRV bump to 1.48.0, https://github.com/steffengy/schannel-rs/issues/75#issuecomment-1117096654 says

> The actual requirement for non-breaking with native-tls and other important dependents probably is something like 1.46.0. As long as that and no other major dependents break, edition / msrv should not matter much.

just FYI, `native-tls` now actually requires Rust 1.53.0: https://github.com/sfackler/rust-native-tls/issues/222

marcograss:
is there any update on this? any blocker? can I help?

thanks

marco

lazka:
Thanks!

PR #87, opened 2022-11-20
update to newest self-signed.badssl.com:443 cert

arlosi:
badssl.com changed their self-signed cert. This PR should fix tests.


steffengy:
Thanks

PR #96, opened 2023-08-30
Added function to set exportable private key flag

cristianbdg:
Hi, I needed the exportable flag to be set and this seems to do the trick. It works on my machine (TM). I would appreciate it a lot if you would consider adding it to the library.

This is my first attempt of a PR. I hope I'm doing it correctly.


steffengy:
Thanks

PR #97, opened 2023-10-11
Add PfxImportOptions::named_no_persist_key()

ironyman:
This function sets PKCS12_NAMED_NO_PERSIST_KEY flag for PFXImportCertStore call to import pkcs12 file to prevent Windows from persisting the keys from the pkcs12 file to disk. Without this flag keys imported from pkcs12 files are saved to disk and need to be deleted manually (they are leaked usually).


steffengy:
Where's PKCS12_NAMED_NO_PERSIST_KEY documented?

PR #114, opened 2025-09-03
Bump windows-sys to 0.60

dpaoliello:


steffengy:
Published with another update (0.61) as 0.1.28.

@djc 
Generally open to that, though generally prefering if there's a track record (previous PRs or other winapi related projects).
Do you have a usecase for it, e.g. personally using/developing projects using it on windows or just by-proxy through users of your own projects? _(So trust wont be an issue here )_
If you're offering, I'll think about it.

djc:
Thanks for publishing the release!

I don't use Windows myself nor do I have a lot of experience with the Windows API. I do maintain a bunch of crates -- for one, I'm one of the rustls maintainers and as such I help maintain rustls-native-certs, which is the most popular dependent for schannel right now. I also maintain chrono and am using windows-bindgen there which helps avoid some of the update churn from the windows-sys crate. If you're generally on top of things here, that's obviously better for me, but if you're too busy/not motivated to keep working on this crate I could help do with mostly passive maintenance (things like dependency bumps and basic code review).

PR #121, opened 2026-03-07
Fix TLS 1.3 data corruption from double-send during renegotiation

nateprewitt:
Let me preface this that I don't regularly develop in Rust so apologies for a naive solution or any idiomatic issues.

---

This is an attempt to address an issue I found while using Reqwest to send large payloads to cloud storage services. When payloads exceed ~1MB, I started seeing TLS Records interspersed in my uploaded data. After digging through the stack I arrived here. I'll try to be succinct with what I found but let me know if you need any extra details.

## Issue
When `read()` and `write()` are interleaved on a non-blocking TLS 1.3 connection (as hyper does in its poll loop), data corruption can occur. The sequence:

1. `write()` encrypts data into `out_buf`, `write_out()` gets `WouldBlock` — encrypted data remains pending in `out_buf`
2. Caller calls `read()` → `fill_buf()` → `read_in()` picks up a TLS 1.3 NewSessionTicket (post-handshake message)
3. `decrypt()` returns `SEC_I_RENEGOTIATE`, transitions state to `Initializing`
4. `fill_buf()` loops → `initialize()` → `write_out()` sends the pending application data from `out_buf`
5. Caller retries `write()` with the same buffer → re-encrypts and sends the same chunk again

The server receives duplicate TLS records, corrupting the payload. This only affects TLS 1.3 because post-handshake messages (NewSessionTicket, KeyUpdate) don't exist in TLS 1.2.

## Proposed Fix
When `SEC_I_RENEGOTIATE` fires and `out_buf` has pending application data, save the current position and advance past it. This prevents `write_out()` inside `initialize()` from sending pending app data during renegotiation. After renegotiation completes, the position is restored so the next `write()` retries normally.

## Reproduction

This issue has been quite hard to reproduce with a localhost server. I'm able to consistently trigger mismatched data uploading >1MB of random data to Azure Blob Storage (or any cloud storage using TLS 1.3) with interleaved `read()`/`write()` on a non-blocking socket. The server's Content-MD5 validation rejects the corrupted upload. Without the fix, uploads fail ~80% of the time. With the fix, I've seen 0 failures across hundreds of test runs.

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -948,3 +948,125 @@ fn test_alpn_list() {
         &full_alpn_list as &[u8]
     );
 }
+
+/// This should reproduce renegotiation error on TLS1.3.
+/// It also verifies the same works on TLS1.2 because why not.
+#[test]
+fn test_renegotiation_corruption() {
+    // This did not reproduce with 2 MB so not sure how reliant this test is.
+    let payload_size = 4 * 1024 * 1024;
+    
+    let mut protos = vec![Protocol::Tls12];
+    if std::env::var("SCHANNEL_SKIP_TLS_13_TEST") != Ok("1".to_owned()) {
+        protos.push(Protocol::Tls13);
+    }
+
+    for proto in protos {
+        let mut data_to_send = vec![0u8; payload_size];
+        let pattern = [0,1,2,3,4];
+        for (i, byte) in data_to_send.iter_mut().enumerate() {
+            *byte = pattern[i % pattern.len()];
+        }
+
+        let cert = match localhost_cert() {
+            Some(cert) => cert,
+            None => return,
+        };
+
+        // 1. Setup TCP on random port
+        let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+        let addr = listener.local_addr().unwrap();
+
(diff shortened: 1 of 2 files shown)


steffengy:
Atleast without digging deeper I do not see the obvious difference between 
write_out doing nothing (flag based if `out_pos < out_len` during reneg) and
write_out doing nothing (because buffer is artificially empty if `out_pos < out_len` during reneg)? 

>  so the pointer is back to it's saved position before any additional data can hit the buffer.

**To summarize the point: Why is it necessary to store the old position, if the buffer wont and cant be changed without causing corruption?**

I realize my wording of `a is_renegotiating flag instead` can be interpreted too broad, e.g. doing it on any renegotiation even if the `out_pos < out_len` condition here is not met.

Nevertheless just wanting to make sure we break as little as possible once rolling out and testing in the wild.
I assume you've tested this with significant TLS1.2/1.3 traffic?

@nateprewitt Also, I've added a reproduction test case in another branch and ensured it fails CI without your commit and passes with. So I think I (& Gemini) got it eventually.
**Can you take a look at that test and if you think its good add it to this PR?**

PR #122, opened 2026-08-31
Add read-only current-user certificate store access

netaneld122:
## Summary
- add a read-only current-user certificate store API
- preserve the existing writable API and share the store-opening implementation
- request `CERT_STORE_READONLY_FLAG` for certificate enumeration without write access

## Test plan
- [x] `cargo check --lib`
- [x] compile `rustls-native-certs` using the new API


netaneld122:
Seems like the test certificate expired on August 12, 2026 😅

steffengy:
Tests are fixed. 
Not wanting this kind of API, since it isnt really future proof, e.g. if another flag comes along need `open_current_user_read_only_and_other_flag`. 
Essentially I would see keeping the basic `open_current_user` etc. we already have and anything beyond that you can configure yourself using a `CertStoreBuilder` or something like that, e.g.
```
let cert_store = CertStoreBuilder::readonly().xyz().current_user()
```

netaneld122:
Very fair suggestion, trying again:
* https://github.com/steffengy/schannel-rs/pull/123

